PR #8, opened 2013-11-01
CB-5235 Resolved InAppBroswer failed test case on WP8

pu-chen:
https://issues.apache.org/jira/browse/CB-5235


surajpindoria:

PR #22, opened 2014-02-17
window.open failing without options parameter

hypermurea:
Documentation indicates options parameter for window.open() is optional, but omitting the parameter on WP8 crashes calls to window.open. Has regressed, so may break existing projects that depend on this plugin.


surajpindoria:

PR #29, opened 2014-03-09
Add a hardwareback option to allow for the hardware back button to go back

kriserickson:
Add's another android option 'hardwareback' that when set allows the android hardware back button to work as a back button in the InAppBrowser until there is no way to go back, at which point it closes the InAppBrowser.


lanash01:
Very much needed change.

purplecabbage:
Thanks Kris, finally got this in!
I changed the default to 'yes' and updated the docs.

Please consider signing the ICLA. Details here: http://cordova.apache.org/#contribute

PR #43, opened 2014-05-06
Added handeling of tel: and mailto: URLs in WP8

Jonasei:


surajpindoria:

PR #50, opened 2014-06-17
Fix opening of files in Windows Phone 8.

henrykuijpers:
The files weren't being opened, because of UriKind.Absolute.

For example: entry.toNativeURL() results in "x-wmapp0://example.pdf/".
Using the existing code, "pathUri.AbsolutePath" results in "/" (because that is the path of the above URL).
And that is definitely NOT the URL we wanted to open...

The code now checks this, reformats the URL and eventually uses the supplied URL to open the file.

### DIFF
--- a/src/wp/InAppBrowser.cs
+++ b/src/wp/InAppBrowser.cs
@@ -14,6 +14,7 @@
 
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


surajpindoria:

PR #62, opened 2014-09-16
Send 'Activity not found' as an error result to Javascript

rromerogar:
When opened an URI with window.open(url,'_system') and throws a "Not Activity found", event 'errorloading' it's not fired. With this request will resolve this.


infil00p:
Closing this PR since it's got too many random commits on it as @clelland has mentioned earlier.

PR #75, opened 2014-12-15
On WP 8.0 `execScript` need to be used instead of `eval`.

durkiewicz:
Executing `browser.InvokeScript("eval", new string[] { anyJsExpression })` on WP 8.0 ends up with an exception.


surajpindoria:

PR #81, opened 2015-02-14
CB 8467: provide hide() method to hide browser window without closing connection

herrevilkitten:
Added support for hiding the web view container.  This maintains the browser
    session without closing it.  The browser window can be repeatedly hidden and
    shown.
    
    ** This has only been tested on android and ios **
    
    amazon/android:
    An additional `hide` action was added to `InAppBrowser#execute`.  It is
    identical to `show`, except that it calls `dialog.hide()` instead.
    
    blackberry10:
    no changes
    
    firefoxos:
    Added a `hide` method that is identical to `show`, indicating it is not
    supported.
    
    ios:
    Added a `hide` method that is identical to `show`, except that it uses
    `dismissViewControllerAnimated`.  It checks the value of
    `_previousStatusBarStyle`.  If it is `-1`, the method returns with no
    action performed.  If it is not, it is set to `-1.`
    
    ubuntu:
    Added a `hide` method that sets `CordovaWrapper.global.inappbrowser.visible` to
    `false`.
    
    windows:
    Added a `hide` method that sets `browserWrap.style.display` to `none`.
    
    wp:
    Added a `hide` method that is identical to `show`, except that it sets
    `browser.Visibility` to `Visibility.Collapsed` and sets `AppBar.IsVisible` to
    `false`.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -152,6 +152,7 @@ The object returned from a call to `window.open`.
 - removeEventListener
 - close
 - show
+- hide
 - executeScript
 - insertCSS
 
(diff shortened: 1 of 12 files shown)


herrevilkitten:
@stevengill Can you review these changes?  I have only used it in iOS and Android.  I presume the Amazon changes will work fine.  For the others, I simply took the "show" method and changed it around in ways I thought were appropriate.

PR #82, opened 2015-02-25
CB-8534: Add support for auth dialogs in inappbrowser plugin

cjpearson:
When an authentication challenge is received, the inappbrowser will now check to see if a plugin can respond to the challenge before going to the default behavior. This is the same behavior that was added to cordova-android in CB-8201. It allows plugins like [Auth Dialog](https://github.com/MSOpenTech/cordova-plugin-auth-dialog) to work with the inappbrowser.

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -41,6 +41,7 @@ import android.view.WindowManager.LayoutParams;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputMethodManager;
 import android.webkit.CookieManager;
+import android.webkit.HttpAuthHandler;
 import android.webkit.WebSettings;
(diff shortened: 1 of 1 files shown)


adrianchia:
Hi,

Is this pull request going to be in the cordova inappbrowser plugin soon?

helzgate:
You could use cordova-plugin-auth-dialog with the webview instead on inAppBrowser.  I just had to remove about 4 lines in a source xml file that was causing issues while building.  Other than that it works great and i'm loving it.

sgrebnov:
This change is required so that cordova-plugin-auth-dialog can work. I'll verify this change and merge

sgrebnov:
Verified and merged

hmusunuru:
@sgrebnov will this work for both ios and android

PR #83, opened 2015-02-25
CB-8438 cordova-plugin-inappbrowser documentation translation

sosahvictor:
CB-8438 cordova-plugin-inappbrowser documentation translation: cordova-plugin-inappbrowser


sosahvictor:
This will not be merged here

PR #85, opened 2015-03-03
Inappbrowser always on top wp8.1

VeldMuijz:
For windows 8.1 the inappbrowser can be displayed with a lower zindex
than the current viewing page. Causing the inappbrowser to be opened on
the background.


purplecabbage:
Handled in #102

PR #89, opened 2015-03-05
CB-3750: Add loading spinner for ios

cjpearson:


syndbg:
I suggest that this is a bit more flexible.

Don't force the color and spinner to be always visible.

Instead introduce two options ex. `spinnerColor` and `spinnerHidden`

sgrebnov:
Tested and merged

ozeranjo:
The ability to pass options for overriding the spinnerColor and hiding the spinner were not included as part of the commit above. Can this be considered for a future release? A valid use case exists if a custom loading progress bar is used -- showing the standard activity indicator can be duplicative and visually confusing.

karlachavez:
Hi guys,

I'm agree with @ozeranjo. At least a hide spinner feature is needed in order to avoid redundance on the site if it already has a custom activity indicator. Is there any plan to include is soon?

shazron:
Please file feature request issues for the needed features at issues.cordova.io
Thanks!

NiklasMerz:
For hide spinner see #254 and [CB-13659](https://issues.apache.org/jira/browse/CB-13659)

PR #90, opened 2015-03-07
Update CDVInAppBrowser.m

byrning:
The current progress bar is white on white background and the spinner frame is incorrect, so it's not visible. The suggested code changes progress bar to gray and adjusts spinner frame.


sgrebnov:
Thank you for your patch, I've just merged similar fix since it was here before (#89), please see if spinner now works fine and consider closing this PR

PR #91, opened 2015-03-08
fixed a bug where status bar disapears

vzaidman:
in case the browser opens hidden and then closed with ref.close().


purplecabbage:
This was fixed in #72

PR #92, opened 2015-03-13
Make toolbar more iOS7 & 8 like

lukeharback:
iOS only changes:
Change the UIToolbar to be full height with no offset to be more like a UINavigationBar
Change the bar style from UIBarStyleBlackOpaque to UIBarStyleBlack. If the webview is not full height the opaque is not being used. Also then setting the translucency to NO
Change the status bar text color to white. Because its a black navigation bar the status bar text should really be white

### DIFF
--- a/src/ios/CDVInAppBrowser.m
+++ b/src/ios/CDVInAppBrowser.m
@@ -28,9 +28,10 @@
 #define    kInAppBrowserToolbarBarPositionBottom @"bottom"
 #define    kInAppBrowserToolbarBarPositionTop @"top"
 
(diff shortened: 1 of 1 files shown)


cjpearson:
I'm seeing some extra grey area when hiding the location bar and showing the toolbar on top.
`window.open('http://google.com', '_blank', 'location=no,toolbarposition=top')`


ios simulator screen shot mar 14 2015 9 29 11 am

Changing TOOLBAR_HEIGHT to TOOLBAR_MARGIN_HEIGHT on line 649 fixes this, but may open some other issues.

PR #99, opened 2015-03-26
inAppBrowser custom application scheme handling on android 

osvso:
Currently the inAppBrowser doesn't handle custom application scheme making it impossible to redirect back to the nativeCode from the embedded webView. The system browser handles such redirects without a problem, but due to security reasons in some cases usage of embedded view is more advisable.

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -57,6 +57,7 @@ import org.apache.cordova.CordovaWebView;
 import org.apache.cordova.LOG;
 import org.apache.cordova.PluginManager;
 import org.apache.cordova.PluginResult;
+import org.apache.cordova.CordovaPreferences;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -83,6 +84,11 @@ public class InAppBrowser extends CordovaPlugin {
     private static final String CLEAR_ALL_CACHE = "clearcache";
     private static final String CLEAR_SESSION_CACHE = "clearsessioncache";
     private static final String HARDWARE_BACK_BUTTON = "hardwareback";
+
+    /*
+    Configuration property. Custom applicaiton scheme to be handeld in a _system.
+     */
+    private static final String CUSTOM_APPLICATION_SCHEME = "CustomAppScheme";
 
     private InAppBrowserDialog dialog;
     private WebView inAppWebView;
@@ -113,7 +119,7 @@ public class InAppBrowser extends CordovaPlugin {
             }
             final String target = t;
             final HashMap<String, Boolean> features = parseFeature(args.optString(2));
-            
+
             Log.d(LOG_TAG, "target = " + target);
             
             this.cordova.getActivity().runOnUiThread(new Runnable() {
@@ -772,6 +778,7 @@ public class InAppBrowser extends CordovaPlugin {
         @Override
         public void onPageStarted(WebView view, String url,  Bitmap favicon) {
             super.onPageStarted(view, url, favicon);
+            final CordovaPreferences preferences = Config.getPreferences();
             String newloc = "";
             if (url.startsWith("http:") || url.startsWith("https:") || url.startsWith("file:")) {
                 newloc = url;
@@ -826,6 +833,18 @@ public class InAppBrowser extends CordovaPlugin {
                 } catch (android.content.ActivityNotFoundException e) {
                     LOG.e(LOG_TAG, "Error sending sms " + url + ":" + e.toString());
                 }
+            }
+            else if (url.startsWith(preferences.getString(CUSTOM_APPLICATION_SCHEME, "!@#$"))) {
+                try {
+                    final Intent intent = new Intent(Intent.ACTION_VIEW);
+                    final Uri uri = Uri.parse(url);
+                    intent.setData(uri);
+                    intent.addCategory(Intent.CATEGORY_BROWSABLE);
+                    cordova.getActivity().startActivity(intent);
+                } catch (final android.content.ActivityNotFoundException e) {
+                    LOG.e(LOG_TAG, "Error during redirection " + url + ":" + e.toString());
+                }
+                newloc = "";
             }
             else {
                 newloc = "http://" + url;


jcesarmobile:
I can't understand what this changes try to achieve. Looking at the code it launchs an intent to open an external app if the url is defined on the config.xml. But that can be done right now using the _system option

osvso:
Maybe I'm missing something. The problem which we faced with the current inAppBrowser plugin implementation is that it doesn't allow to perform the redirection back to the application using custom scheme. Although the android:scheme is registered the the webView in inAppBrowser returns err_unknown_url_scheme when trying navigate back. It works without any problems on iOS, and also using system browser.

osvso:
We don't want to system browser.

purplecabbage:
It should not redirect, it should send a message to it's parent ( the application webview )
The parent should then dismiss it's child (InAppBrowser).

The parent gets notifications every time the location of the InAppBrowser changes, and can respond based on what the location.href is. Think oauth,  and send the InAppBrower to #token=ACCESS_TOKEN or some other signal that you define.

osvso:
I agree that it is responsibility of the parent to dismiss the inAppBrowser (fixed that), but from the other hand it is also responsibility of the webView to handle the redirect. Listening and interpreting redirections from to top level is possible and would probably work, but I consider it as not safe as it moves handling sensitive redirection to javascript layer which can be easily tampered. Also redirection handling will have to be placed within some android specific statement as it works on iOS.

iampeter:
Could this change be introduced? Seems that on some Android devices the "_system" parameter doesn't help with the custom URL scheme.

caphun:
I'm hitting the same problem with a custom scheme that is not on a page I control so cannot add the "_system" target. This fix seems to solve the problem. Are there plans to integrate this soon? If not what is the alternative?

NGumby:
I think a better approach to communicate between an external website and the app, is to raise an event in shouldOverrideLoadUrl like this :

```
public boolean shouldOverrideUrlLoading(WebView webView, String url) {
            LOG.d(LOG_TAG, "shouldOverrideUrlLoading:" + url);

            if (url.startsWith(WebView.SCHEME_TEL)) {
                ...
            }
            else if (url.startsWith("myscheme:")) {
				try {
					JSONObject obj = new JSONObject();
					obj.put("type", "customevent");
					obj.put("url", url);
					sendUpdate(obj, true);
	                return true;
				} catch (JSONException ex) {
					LOG.e(LOG_TAG, "URI passed in has caused a JSON error.");
				}
            }
            return false;
        }
```

add this to InAppBrowser.js 

```
    function InAppBrowser() {
        this.channels = {
            'loadstart': channel.create('loadstart'),
            'loadstop': channel.create('loadstop'),
            'loaderror': channel.create('loaderror'),
            'customevent': channel.create('customevent'),
            'exit': channel.create('exit')
        };
    }

```
in your app, you can then do
              `  appBrowser.addEventListener("customevent", function(event){console.log("My custom event raised:" + event.url)});
`

from the external website, you can then have a link like

`<a href="myscheme://mycustomaction?someparam=value">Custom Action in the app</a>`

It works pretty well, and you can parse the url to do the custom actions!

JPustkuchen:
I can confirm that this is really an important feature request. For example we want to use skype:// and whatsapp:// links within the webview. The problem is, that loadstart acts async and so even a workaround in the event to prevent loading the protocol as page is not possible.

A possible and safe solution (from my point of view) could be a new "beforeloadstart" event which allows to return false and take the url parameter for custom handling. The url should not be prefixed by http:// like currently in loadstart. So we can introduce our own custom logic in JS without hacking the plugin.

What do you think?

osvso:
I've changed the implementation a bit. Currently defining additional property with custom application scheme is not required anymore.

moneytree-doug:
I think this is an important feature for OAuth, has anyone figured out a work around?

ljcljc:
I have met with the same problem, has anyone fix this issue yet?

moneytree-doug:
@ljcljc I used @osvso 's library and it worked.

ljcljc:
@moneytree-doug Thanks for your reply.

Could you explain in detail how it works?

This is the link of @osvso [https://github.com/apache/cordova-plugin-inappbrowser/pull/99/commits/01968d06004645fb6289d35c1ba31eb048c8620d](url)

And, do I need to specify 'CustomSchema' in config.xml? Or it works automatically without any further configuration?

Thanks

ljcljc:
@moneytree-doug I finally figure it out. It works great. No other configurations needed.
@osvso Thanks for your excellent work! Save my day!

JPustkuchen:
@ljcljc : Could you please paste your final solution as example for all of us? That would help a lot :) Thank you!

ljcljc:
@JPustkuchen 
Sure. You can checkout my repository, which I forked from @osvso , using cordova to install it.
[https://github.com/ljcljc/cordova-plugin-inappbrowser](url)

This is his change to the original version.
[https://github.com/apache/cordova-plugin-inappbrowser/pull/99/commits/01968d06004645fb6289d35c1ba31eb048c8620d](url)

Then, you can launch custom scheme within inappbrowser, and that should work fine. No other configuration needed. 
Let me know if you have any questions using it.

moneytree-doug:
@JPustkuchen For example, I forked off the library so to use it as a plugin for cordova then you would do this: ```cordova plugin add https://github.com/moneytree/cordova-plugin-inappbrowser.git```

benbotto:
Seconding @moneytree-doug's statement that this is important for OAuth, or any protocol that requires redirects back to the app.

@ljcljc's fork corrects the problem, and we're using that in our Android builds.  That version doesn't work correctly in iOS 10.2, however, so now we have custom builds for Android and iOS.

Using InAppBrowser, Opening a URL with a custom scheme works fine in iOS.  Please make it work in Android, too!

steinwelberg:
We merged the original master branch into our branch again to make it completely up-to-date. I see that there we introduced other changes which should maybe not be included into this pull request, If that is what is keeping it from being merged I'm happy to create a new pull request with just the changes required to make the in app browser trigger the intent to handle a custom app scheme.

cordova-qa:
Cordova CI Build has one or more failures. 

**Commit**     - [Link](https://github.com/apache/cordova-plugin-inappbrowser/pull/99/commits/e36829ad7042d1dd389b4bc9af6b67c8b306c48e)
**Dashboard** - [Link](http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135/)

| Builder Name  | Console Output | Test Report | Device Logs  |
|     :---:     |     :---:      |   :---:     |     :---:    |
| [Windows 10  Store]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=windows-10-store/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=windows-10-store/console) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=windows-10-store/testReport/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=windows-10-store/artifact/) |
| [iOS 9.3]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-9.3/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-9.3/console) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-9.3/testReport/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-9.3/artifact/) |
| [iOS 10.0]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-10.0/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-10.0/console) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-10.0/testReport/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=ios-10.0/artifact/) |
| [Android 4.4]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-4.4/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-4.4/console) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-4.4/testReport/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-4.4/artifact/) |
| [Android 5.1]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-5.1/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-5.1/console) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-5.1/testReport/) | [Link]( http://cordova-ci.cloudapp.net:8080/job/cordova-plugin-inappbrowser-pr/135//PLATFORM=android-5.1/artifact/) |

benbotto:
Any plan on accepting this pull request?  This issue is more than two years old now, and there's still an inconsistency between iOS and Android that needs to be worked around.

purplecabbage:
Attn: @shazron @infil00p

src/ios/CDVInAppBrowser.m
```diff
@@ -783,6 +793,16 @@ - (void)showToolBar:(BOOL)show : (NSString *) toolbarPosition
...
+    CGRect frame = [UIApplication sharedApplication].statusBarFrame;
+    UIToolbar* bgToolbar = [[UIToolbar alloc] initWithFrame:frame];
+    bgToolbar.barStyle = UIBarStyleDefault;
+    bgToolbar.tintColor = [UIColor whiteColor];
+    bgToolbar.barTintColor = [UIColor whiteColor];
+//    [[UIBarItem appearance] setTitleTextAttributes:@{UITextAttributeTextColor : [UIColor whiteColor]}];
+
+
+    [self.view addSubview:bgToolbar];
+
```
shazron: What's the purpose of this code?

shazron:
From what I have read, it seems this is an Android problem -- so I'm not sure why the code changes in iOS are needed?

src/ios/CDVInAppBrowser.m
steinwelberg: I honestly don't know. That's why I mentioned that this merge request somehow contains more stuff and I'm happy to create a 'clean' merge request.

wilsolutions:
Hi, please, can we get this in?

Kaelink:
That might be interesting for some of you but, while waiting for the pull to be merged, I found a fork that solved the issue here: https://github.com/Onegini/cordova-plugin-inappbrowser

src/ios/CDVInAppBrowser.m
varunkumarm: @steinwelberg Can you please create the new merge request to get the essential part in? Clearly it looks like the extra code is causing the delay. It really helps to have the changes in the official branch than to fork and use.

infil00p:
The code changes in iOS are delaying this.  That being said, I don't know if blindly adding Application URIs in an InAppBrowser is a good idea in this case.

benbotto:
It's really been a long time waiting for this, over two years now!  My team has been manually patching this piece of code to support custom application URIs for over a year.  I'm not familiar enough with the Cordova code or project management to know why the iOS code changes are in there, or if there is a related JIRA ticket.  That said, it sure is inconvenient that custom URIs work differently between iOS and Android.  It seems like either both should work with custom URI schemes, or neither should (my vote's of course for the former).

Anyway, if adding application URIs in an InAppBrowser is not a good idea, then maybe someone can chime in on how to correctly implement something that requires a redirect back to an app.  For example, we use OpenID Connect as a single sign-on solution.  As part of the OpenID Connect spec the user agent (the mobile app) sends the user to a web page to authenticate.  Then, on successful authentication, the user is redirected back to the client with an access token and id token.  That redirect back requires handling a custom URI scheme, and it does not work in Android.

There's a long list of folks affected by this, and it's a burden for your users to have to use custom forks to implement something as common and simple as OAuth2.  That's especially true since the solution, at least ostensibly, seems trivial.  I for one would really appreciate a fix!

purplecabbage:
I blogged about this, and was able to do OAuth2 on iOS and Android without a need for protocols, or applications schemes.  
https://medium.com/@purplecabbage/oauth2-in-apache-cordova-3a3ba059b184

infil00p:
Years ago, we added the Intent whitelist to Android to prevent Android from launching third-party intents that aren't whitelisted.

https://www.cvedetails.com/cve/CVE-2014-3500/

Since the whole point of the InAppBrowser as it was originally intended to be used is to allow untrusted Web content to be run in a sandbox away from the Cordova API, therefore we have to treat the contents themselves as untrusted.  The problem is that there's no way to communicate that for the user except for the address bar, which can be disabled.  It's not clear to the user that they actually made it to the proper OAuth page.  Of course, we don't allow for the InAppBrowser to use Self-Signed Certificates, which does help mitigate this, but we have PRs open for people who want InAppBrowser to do so.

So, if Cordova can't open random intents, neither should the InAppBrowser, and I would actually consider it a bug for iOS to be able to do so.  The InAppBrowser as it is currently designed could be prone to phishing attacks if we start giving it more and more functionality.  Now, if someone was to make this robust enough to check the intent whitelist to see if Cordova could actually launch this intent, I might consider adopting this, but currently there's no way I'd accept this PR as it is.

timbru31:
While I totally agree with you @infil00p, I'd like to point out that Chrome on Android on the other hand *is* able to launch any intent. It blocks custom url schemes like `mycoolapp://`, but allows the intent syntax of it to launch the app: `intent://myIntent/#Intent;scheme=mycoolapp;package=com.mypackage;end`
Maybe this is the direction forward?

For more information refer to this Chrome article: https://developer.chrome.com/multidevice/android/intents

PR #102, opened 2015-04-10
Fixed zIndex issue on Windows 8, 8.1 where InAppBrowser opens behind def...

aroberson:
...ault app.

Windows 8 and 8.1 uses the www/windows8/InAppBrowserProxy.js file instead of the default.  The zIndex on the browserWrap was not being set which causes the window to position behind the app calling window.open().

### DIFF
--- a/www/windows8/InAppBrowserProxy.js
+++ b/www/windows8/InAppBrowserProxy.js
@@ -62,6 +62,7 @@ var IAB = {
                 browserWrap.style.borderWidth = "40px";
                 browserWrap.style.borderStyle = "solid";
                 browserWrap.style.borderColor = "rgba(0,0,0,0.25)";
+                browserWrap.style.zIndex = "9999999";
 
                 browserWrap.onclick = function () {
                     setTimeout(function () {
--- a/www/windows8/InAppBrowserProxy.js
+++ b/www/windows8/InAppBrowserProxy.js
@@ -17,11 +17,10 @@
  * specific language governing permissions and limitations
  * under the License.
  *
-*/
+ */
 
 /*jslint sloppy:true */
 /*global Windows:true, require, document, setTimeout, window, module */
-
 
 
 var cordova = require('cordova'),
@@ -39,9 +38,9 @@ var IAB = {
     },
     show: function (win, lose) {
         /* empty block, ran out of bacon?
-        if (browserWrap) {
+         if (browserWrap) {
 
-        }*/
+         }*/
     },
     open: function (win, lose, args) {
         var strUrl = args[0],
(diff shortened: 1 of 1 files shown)


icaro56:
I follow your sample. Now the window shows. But, it does not scale in screen. See the picture.

inappbrowsertilebug

aroberson:
You have to use css and divs to get it to position in the web view or iframe properly.

icaro56:
aroberson,

Have I change the InAppBrowserProxy.js file to solved this?

aroberson:
No, after making the zindex change, you have have your apps css right for the content displayed.

icaro56:
On Windows 10 it does not works. The InAppBrowserProxy.js never is called. Could you help me?

icaro56:
In inappbrowser's plugin.xml the src for platform Windows src/windows/InAppBrowserProxy.js does not exist. Then I change to src/windows8/InAppBrowserProxy.js.

icaro56:
facebook

Now, it is like this.

The css files from facebook are not working properly.

SEC7117: Network request to https://fbstatic-a.akamaihd.net/rsrc.php/v2/yX/r/qVownUs0hjp.css did not succeed.
Your application manifest does not declare the following capabilities: internetClient privateNetworkClientServer

aroberson:
There is nothing I can do without seeing the code.  It would take a lot of time to understand what you are doing.

icaro56:
aroberson, try to open this link on windows 10 emulator with window.open

https://m.facebook.com/login.php?skip_api_login=1&api_key=321311011387171&signed_next=1&next=https%3A%2F%2Fm.facebook.com%2Fv2.2%2Fdialog%2Foauth%3Fredirect_uri%3Dhttps%253A%252F%252Fwww.facebook.com%252Fconnect%252Flogin_success.html%26scope%3Demail%252Cuser_friends%252Cpublic_profile%26response_type%3Dtoken%26client_id%3D321311011387171%26ret%3Dlogin&cancel_url=https%3A%2F%2Fwww.facebook.com%2Fconnect%2Flogin_success.html%3Ferror%3Daccess_denied%26error_code%3D200%26error_description%3DPermissions%2Berror%26error_reason%3Duser_denied%23_%3D_&display=touch&_rdr

icaro56:
Aroberson,

I send a email with the code.

thanks.

napcat:
Is this fixed? How can I fixe it?
I have the exact same problem described in the image above:
https://cloud.githubusercontent.com/assets/7433049/8339931/80bbcee4-1a91-11e5-8fe1-c6112debd384.png

I use inAppBrowser to login on facebook, google or twitter in my login screen. The inAppBrowser appears, but not fully scaled.

I'm using cordova 5.3.3 and cordova-plugin-inappbrowser 1.0.1 "InAppBrowser".
The same code works fine on a Windows Phone 8 platform (cordova add platform wp8), but not in windows platform (cordova add platform windows), the projects for windows phone 8.1 and windows 10 (universal windows) does not work as aspected (not fully scaled).

icaro56:
Do the test on windows 10 device. It seems the emulator is broken.

eduboxgithub:
I'm not testing in the emulator... I have an old Lumia 625 with windows 8.1 and a Lumia 640 with windows 10. In both devices the webview is not fully scaled and the image above.

If I use wp8 cordova platform it works great in both devices, but I need to upgrade the app to 8.1 at least  =(

Can it be some weird css in my project? I don't think I have any specific css for the webview, at least I did not add any...

napcat:
I'm not testing in the emulator... I have an old Lumia 625 with windows 8.1 and a Lumia 640 with windows 10. In both devices the webview is not fully scaled and the image above.

If I use wp8 cordova platform it works great in both devices, but I need to upgrade the app to 8.1 at least =(

Can it be some weird css in my project? I don't think I have any specific css for the webview, at least I did not add any...

sgrebnov:
There are two issues here IMO:
1. The fix goes to `src/windows8`- I think we can just use `src/windows/` for `windows8` platform as well - I'll test and update `plugin.xml`
2. I think we should use `box-sizing: border-box` and `width:100%` so we don't need to calculate new width when size is changed

I'll handle this

napcat:
thank you sgrebnov for your time!

let me know if you found a fix for it.

sgrebnov:
@napcat please try the latest plugin version and let me know whether it works
```
cordova plugin add https://github.com/apache/cordova-plugin-inappbrowser/
```

I've been playing with re-size and it worked for me, I don't see anything wrong with InAppBrowser popup size (`box-sizing: border-box` and 100% width have been recently added to master)

napcat:
Yes... that worked! =) The inappbrowser windows is now fully scaled, but there's a border, with about 70/80px around the page. Is that supposed to happen?

Another thing... If i press the back button on the device, instead of going back to the app and close the in app browser, the hole app closes.

purplecabbage:
The 80px border is by design. I am planning a pr to make this customizable via css. 
Back button is an error, I believe there is a jira issue already.

napcat:
Hello, purplecabbage. Thank you for the information.

I changed the css to remove the border, but I'm assuming the border serves as a way to close the in app browser (if I click in the empty area the inappbrowser closes), since there's a bug with the back button. I will wait for that fix.

thank you.

sgrebnov:
I believe we have special option (documented) for this case  (fullscreen=yes)  and two classes you can use to customize IAB popup look and feel (`inAppBrowserWrap` and `inAppBrowserWrapFullscreen`)
https://github.com/apache/cordova-plugin-inappbrowser/blob/master/www/inappbrowser.css

PR #104, opened 2015-06-11
 CB-9158 - InAppBrowser zoomControls are always set to true

syndbg:
Also minor whitespaces fixes.

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -91,9 +91,9 @@ public class InAppBrowser extends CordovaPlugin {
     private boolean showLocationBar = true;
     private boolean showZoomControls = true;
     private boolean openWindowHidden = false;
-    private boolean clearAllCache= false;
-    private boolean clearSessionCache=false;
-    private boolean hadwareBackButton=true;
+    private boolean clearAllCache = false;
+    private boolean clearSessionCache = false;
+    private boolean hadwareBackButton = true;
 
     /**
      * Executes the request and returns PluginResult.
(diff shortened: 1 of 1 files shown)


src/android/InAppBrowser.java
```diff
@@ -656,7 +647,7 @@ public void onClick(View v) {
                 WebSettings settings = inAppWebView.getSettings();
                 settings.setJavaScriptEnabled(true);
                 settings.setJavaScriptCanOpenWindowsAutomatically(true);
-                settings.setBuiltInZoomControls(getShowZoomControls());
+                settings.setBuiltInZoomControls(showZoomControls);
```
syndbg: here's the issue

infil00p:
Why does this pull request touch iOS?  Can you make it so that it's only Android?

purplecabbage:
Numerous issues here .. removing whitespace from the diff reveals only minor changes.
The changes however do not address the issue as reported in JIRA.

syndbg:
The white spaces are automatically stripped by my text editor.

syndbg:
@purplecabbage Good catch. In the JIRA ticket, unfortunately I've pasted the wrong option parameter, `location`, when in fact I meant `zoom`.

This PR fixes `zoom=true` or `zoom=false` not making a difference in the android app.

Can we agree on the fact that fixing (in my opinion) non-formatted code, even in the iOS section of the plugin, should not be a blocker to merge this PR.

purplecabbage:
It is a blocker typically.  Modifications should be restricted to resolving the issue at hand, or at the bare minimum whitespace/formatting changes should be in their own commit.
Reviewing pull requests is enough work without the overhead of having to manually sort through what is relevant.

infil00p:
Agree with @purplecabbage, the iOS whitespace stripping is a blocker.

syndbg:
Alright. Sorry for the delays guys.

Rebased latest upstream master, didn't commit the iOS file and updated the commit message to make more sense.

purplecabbage:
I have tested the change, and verified it. So it is now merged.
Thanks for the contribution @syndbg 
Some great next steps would be to document the feature (under android quirks) and to add a manual test to the list in the /tests/tests.js file.

PR #109, opened 2015-07-16
CB-9378 Fix InAppBrowser not taking whole screen on Windows

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-9378)


dblotsky:
LGTM

PR #111, opened 2015-07-21
give the ability to trust a certificate authority

brad:
We found ourselves in a position where we needed to access a site with a certificate issued by a CA not trusted by default in Android. In this case the browser fails silently and remains blank. This is the workaround we came up with:
- You save the CA certificate to `www/trusted-ca.der` when building your cordova app.
- When the plugin loads an https site, the CA gets added to the trusted CAs for `HttpsUrlConnection`
- When an `SSLError` is received, we do the following:
  - Check if the domain of the URL the error is about is in the whitelist (which starts out empty). If so, we ignore the error and proceed.
  - If not, we check if `HttpsUrlConnection` can connect to the domain without a problem. If so, we add it to the whitelist, ignore the error and proceed. If not, we fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,13 @@ simply hook `window.open` during initialization.  For example:
         window.open = cordova.InAppBrowser.open;
     }
 
+If you need to access any sites that use certificates which aren't normally
+trusted, such as self-signed certificates. You can force the InAppBrowser to
+trust them by saving the CA certificate (in DER format) to a file called
+`trusted-ca.der` in your `www` folder. The InAppBrowser will load the file and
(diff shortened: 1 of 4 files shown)


infil00p:
This has the following wrong with it:
1. You changed the package.json
2. You don't have a JIRA issue attached to this
3. There are other conflicts with the current InAppBrowser code.

If you are still interested in contributing, please fix these issues, and we can then take a look at this.

brad:
@infil00p Sorry about that. I will make a new PR without those issues

brad:
Superseded by https://github.com/apache/cordova-plugin-inappbrowser/pull/243

PR #116, opened 2015-09-22
Added support for file:/// access

bpresles:
I had the requirement to be able to access file:/// image on my inappbrowser popup in Android. But I had issues with Security.

This was because the WebView wasn't configured to allow universal access from file URLs (required for Jelly Bean or greater).

I think this would be a good idea to include this change in the original plugin, as I'm sure I'm not the only one facing this issue.

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -650,6 +650,10 @@ public class InAppBrowser extends CordovaPlugin {
                 settings.setBuiltInZoomControls(showZoomControls);
                 settings.setPluginState(android.webkit.WebSettings.PluginState.ON);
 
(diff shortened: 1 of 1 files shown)


mani95lisa:
+1

shazron:
Thanks for this PR. There's 3 issues with this PR:
1. It doesn't merge cleanly anymore - the commits should be rebased
2. An issue should be created for this in issues.cordova.io so it can be tracked (as you can see there are lots of PRs here that are less likely to be tracked vs JIRA)
3. Once an issue is created, prefix the PR with the issue token, e.g. "CB-XXXX" so it can be automatically tracked to the issue

infil00p:
Closing this PR as abandoned, since it's been over a year since @shazron last checked in on this.

PR #118, opened 2015-10-07
CB-9760 InAppBrowser: fallback to default window.open behavior on Ripple

sgrebnov:
https://issues.apache.org/jira/browse/CB-9760


vladimir-kotikov:
LGTM

PR #121, opened 2015-10-27
Fixed issues causing spinner to not show on web load.

BenHarnett:
As the spinner and the background were white the spinner was not
showing. Also the frame was being set without taking into consideration
of the site of the main view.


sgrebnov:
Hey Ben, thank you for your patch - it looks good. I've just merged similar fix that was here before (#89) so please take a look whether spinner works for you in master and consider closing this pr

jcesarmobile:
Closed the PR, it was already covered by #89

PR #122, opened 2015-10-29
Added Android cookie sharing functionality between webviews

Sinistralis:
There currently exists a gap in functionality with Android as opposed to iOS where iOS will actively share cookies with any inappbrowser webviews, but Android does not. This simply adds in a hook that allows Android to share or get cookies between the Cordova webview and the inappbrowser webview.

Please see the existing discussion below and contribute to it if there are any concerns.

### DIFF
--- a/src/amazon/InAppBrowser.java
+++ b/src/amazon/InAppBrowser.java
@@ -164,6 +164,26 @@ public class InAppBrowser extends CordovaPlugin {
                 }
             });
         }
+        else if (action.equals("getCookies")) {
+            String cookieUrl = args.getString(0);
+            String[] cookies = CookieManager.getInstance().getCookie(cookieUrl).split(";");
+            for (String cookie: cookies) {
+                webView.getCookieManager().setCookie(cookieUrl, cookie);
+            }
+            PluginResult pluginResult = new PluginResult(PluginResult.Status.OK, cookies.length);
+            pluginResult.setKeepCallback(true);
+            callbackContext.sendPluginResult(pluginResult);
+        }
+        else if (action.equals("setCookies")) {
+            String cookieUrl = args.getString(0);
+            String[] cookies = webView.getCookieManager().getCookie(args.getString(0)).split(";");
+            for (String cookie: cookies) {
+                CookieManager.getInstance().setCookie(cookieUrl, cookie);
+            }
+            PluginResult pluginResult = new PluginResult(PluginResult.Status.OK, cookies.length);
+            pluginResult.setKeepCallback(true);
+            callbackContext.sendPluginResult(pluginResult);
+        }
         else if (action.equals("close")) {
             closeDialog();
         }
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -192,6 +192,26 @@ public class InAppBrowser extends CordovaPlugin {
                 }
             });
         }
+        else if (action.equals("getCookies")) {
+            String cookieUrl = args.getString(0);
+            String[] cookies = CookieManager.getInstance().getCookie(cookieUrl).split(";");
+            for (String cookie: cookies) {
+                webView.getCookieManager().setCookie(cookieUrl, cookie);
+            }
+            PluginResult pluginResult = new PluginResult(PluginResult.Status.OK, cookies.length);
+            pluginResult.setKeepCallback(true);
+            callbackContext.sendPluginResult(pluginResult);
+        }
+        else if (action.equals("setCookies")) {
+            String cookieUrl = args.getString(0);
+            String[] cookies = webView.getCookieManager().getCookie(args.getString(0)).split(";");
+            for (String cookie: cookies) {
+                CookieManager.getInstance().setCookie(cookieUrl, cookie);
+            }
+            PluginResult pluginResult = new PluginResult(PluginResult.Status.OK, cookies.length);
+            pluginResult.setKeepCallback(true);
+            callbackContext.sendPluginResult(pluginResult);
+        }
         else if (action.equals("close")) {
             closeDialog();
         }
--- a/www/inappbrowser.js
+++ b/www/inappbrowser.js
@@ -80,6 +80,24 @@ InAppBrowser.prototype = {
         } else {
             throw new Error('insertCSS requires exactly one of code or file to be specified');
         }
+    },
+
+    //Gets cookie by Domain from the inAppWebView and adds it to the cordovaWebView.
+    getCookies: function(cookieDetails, cb) {
+        if(cookieDetails.url) {
+            exec(cb, null, "InAppBrowser", "getCookies", [cookieDetails.url, !!cb])
+        } else {
+            throw new Error('getCookie requires a url to be specified')
+        }
+    },
+
+    //Gets cookie by Domain from the cordovaWebView and adds it to the inAppWebView.
+    setCookies: function(cookieDtails, cb) {
+        if(cookieDetails.url) {
+            exec(cb, null, "InAppBrowser", "setCookies", [cookieDetails.url, !!cb])
+        } else {
+            throw new Error('setCookie requires a url to be specified')
+        }
     }
 };
 
@@ -107,4 +125,3 @@ module.exports = function(strUrl, strWindowName, strWindowFeatures, callbacks) {
     exec(cb, cb, "InAppBrowser", "open", [strUrl, strWindowName, strWindowFeatures]);
     return iab;
 };
-


purplecabbage:
The fact that iOS shares cookies is a defect imho.
That said, I think what 'should' happen is that this 'feature' should be exposed as an option.
I would expect the default behavior to be no-cookie-sharing! The original intent of the iab plugin was to allow non-secure content to be displayed in your app without worrying about security.

Also, do domain rules apply? If so this is less of an issue.
If I load google.com inside the iab on iOS, can google.com access cookies I set from within my app?

Sinistralis:
>The fact that iOS shares cookies is a defect imho.

I would be completely ok with agreeing with this. I just want constancy, assuming...

>That said, I think what 'should' happen is that this 'feature' should be exposed as an option.

the above is the case. Sharing cookies in an HTTP only environment is crucial to certain workflows, assuming the cookie is not passed into the JS code.



>Also, do domain rules apply? If so this is less of an issue. If I load google.com inside the iab on iOS, can google.com access cookies I set from within my app?

Yes it can, this was actually how I was testing this. I needed the tab to log into an SSO workflow, and then on url change start making ajax requests in the cordova webview. Currently on iOS, once the spawned iab completes it's transaction, the cordova webview is able to make secured web calls using the token that the iab received.

purplecabbage:
Thanks for the info.
I think this is actually fine, as long as the options `clearsessioncache` and `clearcache` are honored, as documented.

Sinistralis:
Correct, my thought process was that you would want to absolutely clear these cookies upon closing the app.

Which leads into a question of, should there be a domain argument on these calls?

purplecabbage:
The clear-cache options also clear the cookies in the inappbrowser before a page loads, so if you wanted to load something you didn't trust, you would use that option before the content is loaded to make sure that content/site could not read your cookies.

As far as the cordova-webview accessing cookies set inside inappbrowser by a remote site, I think that is much less of a concern, and possibly even something we would just allow ( much like any native app can do )

Sinistralis:
I think we have different concerns here even though we agree on the same
thing. If the app developer is concerned about exposing their cookies to a
website they should be using HTTP only cookies

nicoabie:
Does this work when using XWalk? Is there any way to make it work using XWalk? Because we are not getting any cookies setted. Thanks in advance

Sinistralis:
When you say you do not get cookies set, are you talking about using this code or default functionality? I did not test against XWalk, but I would not be opposed to trying it out.

nicoabie:
Sorry for not being explicit enough, I made a fork of the inappbrowser code and applied this pull request. I'm using XWalk and debuggind the code I get the cookies for the domain of the IAB but when they are set into the XWalk webview, let's say they are not being accepted, it silently fails

Sinistralis:
I never tried this with XWalk but I'm not opposed to checking it out. I'm
at Rich Web right now so give me a few days to look into this

nicoabie:
Thanks @Sinistralis

refactorthis:
Any news on this one? I also need this functionality. We have a somewhat common scenario of using third party oauth provider in isolated window, currently our return url inside the window will create a cookie, for security this must be http only so sadly do not have access to the cookie from the main app window. 

Will load from the fork for the time being.

infil00p:
Is there work being done on this? It'd probably be good to update the pull request so that it can be more easily tested against the current master.

Sinistralis:
I've been meaning to come back to this. I've been incredibly busying reforming the front-end process at my job. I'm trying to find the time but havn't so far.

Sinistralis:
Ok, so its looking like I can spend some time on this tommorrow (or today really. I'm up way too late). I went ahead and cleaned up the conflict and updated this branch. I will look at adding tests and integrating with crosswalk and hopefully have something for tommorrow.

JoMas971:
Hello @Sinistralis, can you give us an update ?

I also need this for my app, i'm actually using Xwalk (for android development, you can't really avoid it... It is much better in all point), but the cookies with InAppBrowser are not shared and i've some issue when i'm loading secured iFrames containing user informations (so, need to be connected to be able to see the content) but the cookie is not shared and the iframe contains my login page though -_-.
 
Note : Have you already trying to use the new wkwebview for iOS app ? I do, and i have the same issue than android, the cookies is not shared so my secured iFrames ask me to login again and again and again lol.

Thank you for your work by the way, it is amazing !

tanansatpal:
Hello @Sinistralis,
I am also using Xwalk for my android app. It would be a great help if this feature is introduced.

Sinistralis:
When I created this, it was meant to be used as a proof of concept for something at my place of employment. We recently dropped this path so I havn't had the time to come back to this, since as I stated earlier up I am bogged down with ensuring our transition to a React stack is going smoothly. 

I'm happy to look into this further, but its simply a matter of having the time. Past few months have been crazy for me. 

Can someone toss whats needed for me to pull in XWalk? I can try to get around to this again now that it has no conflicts and is moslty up to date.

JoMas971:
Hi @Sinistralis, in fact i've done a fork and merge your PR to test it out. 
It seems that you miss only one thing with your PR : the inappbrowser.js is loaded for iOS and other platforms too. But, since you haven't created the getCookies/setCookies methods for them, the build fail for these platforms and that's not good at all ! (1)
To fix this, i've used the cordova device plugin to execute the `exec(...)` function only on android, but that mean you'll need to have the cordova-device-plugin installed. I don't think it's the best way to do that, and it can definitely be improved but it's still a working solution. (you can see my fork here : https://github.com/ilanddev/cordova-plugin-inappbrowser/tree/mergin_1.3.1).

Otherwise, i'm using crosswalk, and this PR seems to fix my cookies sharing issues. I just have to call manually the getCookies() function before the window.close() action when i log in to my app using Keycloak, to share the cookies inside my popup with my crosswalk webview allowing me to use Keycloak iFrames inside my app. Which is really good :-).

This is the only working way i've found to share cookies between 2 different webviews. The InAppBrowser plugin use the native UIWebView and, it seems, they will not change this behaviour, so we need something to share cookies between 2 different WebViews system (Native WebView and WateverOtherWebView).

The same thing happens if you want to use the iOS WKWebView instead of the UIWebView. Since it is a different WebView system, you can't share cookies between them (it is not automatic). It could be really cool if we can have these magic methods for iOS toà (and that will fix my (1) point !). That way we can use the new and powerful WKWebView for our iOS application too. Have you consider to do some Objective C code ? :-D

Thank you for your work by the way !

Hope this will be merge someday, it is definitely a solution for those who use different WebView system :-).

JPustkuchen:
+1 for this merge. A really useful feature! :+1:

Naveenbc:
Everybody is saying cookies are shared between the web views in IOS. But how to access a cookie that is set in Inappbrowser? Kindly help me in fetching cookies in IOS.
Thanks in advance.

chriscant:
I found that getCookies() was just what I needed for Android crosswalk for the user to log into Yahoo in InAppBrowser and then make the cookies available to the main WebView so ajax calls can be made to Yahoo.

novastorm123:
In IOS 10, Its the opposite. It is working in Android now but not working in iOS10. Can anyone help
setting clearcache and clearsessioncache to no is not working.

ravindrapalli33:
How to manually call get cookies?

danova123:
Unable to call getCookies manually.

infil00p:
So, is this functionality not working? It would help if there was a JIRA issue attached to this PR.  I don't see XWalk functionality being a huge issue since XWalk is no longer being maintained.  Because of that, it's more likely this will get pulled in.  However, it'd be helpful if someone could tell me if this actually works before I get around to testing it.

FiReBlUe45:
Hello,

I would like to share session cookies because I have to login via the default browser and when I go back to my application I launch InAppBrowser with this cookie.

It works with the plugin "browser tab" but I need "InAppAorwser" features to listen to events for example

An idea??

thesourav123:
@FiReBlUe45 
To use it with InAppBrowser, see here https://github.com/apache/cordova-plugin-inappbrowser/pull/238

FiReBlUe45:
@thesourav123 Thank you for your answer but that does not solve my problem, I try to have access to chrome session cookies to copy them after in my application.

I received a proposal for resolution but it requires access to the web application
https://issues.apache.org/jira/browse/CB-11045

janpio:
Hey @Sinistralis, there seems to be a merge conflict now - because of the not any more supported `amazon` platform. Could you maybe fix this by just removing the file? Thanks.

Also, rebasing on `master` should give your PR some tests which would make merging this much easier. Thanks.

PR #126, opened 2015-11-20
Allow for navigateTo method

byrning:
Adding this method allows for one to navigate to another URL while the inappbrowser is already open.

### DIFF
--- a/www/inappbrowser.js
+++ b/www/inappbrowser.js
@@ -45,6 +45,9 @@ InAppBrowser.prototype = {
             this.channels[event.type].fire(event);
         }
     },
+    navigateTo: function (newurl) {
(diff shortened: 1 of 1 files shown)


purplecabbage:
This will not work. There would need to be a native implementation to back this, and it would need to exist in all platforms.

Try this:
`var navigateTo = function(url) {
    return cordova.InAppBrowser.open(url,'_blank');
};`

sgrebnov:
I think we should close this PR as this won't work on other platforms and you can also do the following:
```
ref.executeScript({code: 'location.href="http://some-page.html"'})
```

purplecabbage:
Yeah, this pull-request is invalid, even if the feature request makes sense, this does not do it.
please close!

PR #127, opened 2015-11-25
Fix syntax error when browserifying

alexbuijs:
When browserifying this plugin, aliasify throws an error:
```
events.js:141
      throw er; // Unhandled 'error' event
      ^

SyntaxError: 'return' outside of function (25:4) (while aliasify was processing plugins/cordova-plugin-inappbrowser/www/inappbrowser.js) while parsing file: plugins/cordova-plugin-inappbrowser/www/inappbrowser.js
```

### DIFF
--- a/www/inappbrowser.js
+++ b/www/inappbrowser.js
@@ -22,7 +22,6 @@
 // special patch to correctly work on Ripple emulator (CB-9760)
 if (window.parent && !!window.parent.ripple) { // https://gist.github.com/triceam/4658021
     module.exports = window.open.bind(window); // fallback to default window.open behaviour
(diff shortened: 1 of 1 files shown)


sgrebnov:
@alexbuijs thank you for proposed patch, but we can't remove this line as it is required for InAppBrowser to work correct when app is emulated using Ripple emulator. The problem has been fixed here: https://github.com/apache/cordova-plugin-inappbrowser/commit/2e0dc26eaa023e9d553a78c4acf25af751cc8021?w=1
Please consider closing this PR

alexbuijs:
I understand, thanks!

PR #129, opened 2015-12-17
CB-10009 Improve InAppBrowser toolbar look and feel on Windows

sgrebnov:
https://issues.apache.org/jira/browse/CB-10009

Solution is based on `Segoe UI Symbol` font which is available on all windows devices (tested) and contains required icons.

**Original version:**

appbar-original

**Proposed:**

appbar-disabled

appbar-hower


nikhilkh:
Looks like a good one to merge!

sgrebnov:
Re-based on top of master and merged

PR #132, opened 2015-12-30
CB-7786 android: add new param

seank-com:
- param has same name and default as iOS
- setting is conditionally gated the same way as
  CordovaLib/src/org/apache/cordova/engine/SystemWebViewEngine.java
- updated manual video tests
- updated docs in README.md
- added notes and steps for running plugin tests to CONTRIBUTING.md
- I'm not neuortic, atom removed uneeded whitespace automatically


sgrebnov:
LGTM

PR #133, opened 2016-01-05
handle app store urls in system browser

uareurapid:
I´ve found the following situation (on IOS app)...

I navigate to some website that has links to app store apps...

If the browser finds "https://itunes.apple.com" links they are turned into "itms-appss://" or "itms-apps://" and the browser fails to load them because it doesn´t know to handle them.

These should be opened in system browser and the webview should give up on them, cause it will fail to process them


purplecabbage:
LGTM!

magnusburton:
Looks great!

PR #136, opened 2016-01-15
CB-8180: Changing methods of interception in WebViewClient class

infil00p:


infil00p:
Neat, that actually worked!

falkobuttler:
This PR caused a major outage for our Android users due to the fact that the LOAD_START event does not fire anymore. We relied on that event to move the user to a different flow. I HIGHLY encourage you to revert that part of the change and re-introduce the `loadstart` event

falkobuttler:
@asfgit

jcesarmobile:
@falkobuttler it has been reported https://issues.apache.org/jira/browse/CB-10407
For now you can keep using the old version of the plugin until this is fixed and released

nraboy:
I can confirm this issue as well.  

I've added my two cents to the Jira ticket, but it is causing major problems for my open source library ng-cordova-oauth as referenced in https://github.com/nraboy/ng-cordova-oauth/issues/193

PR #137, opened 2016-01-16
Fix for CB-10305

jcesarmobile:
There are really 3 issues here.
1. When the inAppBrowser is started in portrait and rotates, the
background of the status bar doesn’t fill the whole area. (any iOS
version)
2. On iOS 7, if the inAppBrowser plugin is started on landscape, the
background of the status bar was vertical.
3. On iOS 8-9, on iPhone/iPod, the status bar disappears on landscape,
but the background doesn’t disappear. I’ve changed it to make the
status bar appear  on landscape too. (This wasn’t on CB-10305, but I
think it’s related)

### DIFF
--- a/src/ios/CDVInAppBrowser.m
+++ b/src/ios/CDVInAppBrowser.m
@@ -748,6 +748,10 @@
     return UIStatusBarStyleDefault;
 }
 
+- (BOOL)prefersStatusBarHidden {
+    return NO;
+}
+
 - (void)close
 {
     [CDVUserAgentUtil releaseLock:&_userAgentLockToken];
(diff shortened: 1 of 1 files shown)


sgrebnov:
Is it still actual? - Looks like we have dropped iOS 7.0 support - see CB-9726 Update minimum Deployment Target to iOS 8.0 (https://issues.apache.org/jira/browse/CB-9726)

jcesarmobile:
It fix 2 other things that affect iOS 8-9 too.
And don't we have to maintain support for 6 months after deprecating? (not sure if that applies to plugins too)

sgrebnov:
I see, merging...

PR #139, opened 2016-01-19
Add fix for iOS file upload.

birge:
I'd like to start by saying I don't understand fully what the issue is or how this fixes the issue. I have been having an issue where my cordova app would fall back to the splash screen when I tried to upload a file on iOS. I found the solution here https://issues.apache.org/jira/browse/CB-7679 

This is before the fix
file_upload_bug

and this is after
file_upload_bug_fixed

I am not sure why this fix wasn't already included so I am hoping it just because no one has taken the time to make a PR. Any issues with this?

### DIFF
--- a/src/ios/CDVInAppBrowser.m
+++ b/src/ios/CDVInAppBrowser.m
@@ -987,6 +987,12 @@
 
 @implementation CDVInAppBrowserNavigationController : UINavigationController
 
+- (void) dismissViewControllerAnimated:(BOOL)flag completion:(void (^)(void))completion {
+    if ( self.presentedViewController) {
+        [super dismissViewControllerAnimated:flag completion:completion];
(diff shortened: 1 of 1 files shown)


hkjorgensen:
bump

spedmo:
Why isnt this in?

firepol:
This should be definitively in. I created a fork, modified the file as birge did and installed from my modified repo. Would have been nicer taking this from the official repo...

worsection:
Will you add this fix or any alternative to fix ios file upload?

jcesarmobile:
Thanks for the pull request. I've had a look at it and think it looks good. Before we can merge it though, you need to sign Apache's Contributor License Agreement (can be done online): http://www.apache.org/licenses/#clas

PR #141, opened 2016-01-26
CB-10428 Fix syntax error when browserifying inAppBrowser plugin

sgrebnov:
https://issues.apache.org/jira/browse/CB-10428

Wrapped module in anonymous function

### DIFF
--- a/www/inappbrowser.js
+++ b/www/inappbrowser.js
@@ -19,92 +19,93 @@
  *
 */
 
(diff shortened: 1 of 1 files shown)


sgrebnov:
Use  `?w=1` param to ignore space changes:
https://github.com/apache/cordova-plugin-inappbrowser/pull/141/files?w=1

vladimir-kotikov:
LGTM

PR #142, opened 2016-01-27
CB-10441 Add auto tests for InAppBrowser plugin

sgrebnov:
https://issues.apache.org/jira/browse/CB-10441

Added the following unit tests:
1. `cordova.InAppBrowser.open` method exists and returns `InAppBrowser` instance as a result
2. `loadstart`, `loadstop`, `exit`, `loaderror` events are correctly supported


vladimir-kotikov:
LGTM

PR #143, opened 2016-01-28
CB-10456 InAppBrowser is not closed if I close it programmatically on Android

sgrebnov:
https://issues.apache.org/jira/browse/CB-10456

This happens due to race condition: open/initialization logic is queued into ui thread, but dialog close method test for null is not, so if I do open and then close then close method test is executed before open method is completed.


vladimir-kotikov:
LGTM

PR #144, opened 2016-01-28
CB-6702 InAppBrowser hangs when opening more than one instance

sgrebnov:
https://issues.apache.org/jira/browse/CB-6702


sgrebnov:
This fix is similar to what 	@mysegfault proposed below. I've just added logic to prevent the dialog for being dismissed twice.
https://github.com/apache/cordova-plugin-inappbrowser/pull/71/files

nikhilkh:
LGTM!

PR #145, opened 2016-01-28
InAppBrowser bugfixing on Windows: CB-10451, CB-10452, CB-10454, CB-10450

sgrebnov:
CB-10451 InAppBrowser: loadstart event is not triggered on Windows
CB-10452 InAppBrowser: 'exit' event is not triggered on Windows
CB-10454 InAppBrowser: 'loaderror' event does not have code and message on Windows
CB-10450 InAppBrowser: Unable to get property 'canGoBack' of undefined on Windows

### DIFF
--- a/src/windows/InAppBrowserProxy.js
+++ b/src/windows/InAppBrowserProxy.js
@@ -25,7 +25,6 @@
 
 
 var cordova = require('cordova'),
-    channel = require('cordova/channel'),
     urlutil = require('cordova/urlutil');
 
 var browserWrap,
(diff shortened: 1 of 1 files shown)


sgrebnov:
Use the following url to review proposed changes w/o whitespace: https://github.com/apache/cordova-plugin-inappbrowser/pull/145/files?w=1

nikhilkh:
LGTM. Looks like these bugs have been there for a while and this is not a recent regression that this change is addressing.

nikhilkh:
I'm curious why those `setTimeout()` calls are required.

nikhilkh:
Also, I would recommend using `setImmediate` as it has advantages over `setTimeout(fn, 0)`. https://msdn.microsoft.com/en-us/library/windows/apps/hh453394.aspx

sgrebnov:
setTimeout is required to support the following scenario. On ios, android and other platforms functionality creates InAppBrowser on native side, so ln2 happens before webview is created and events are triggered. As a quick fix on Windows I made open method async so that it works (this is what docs recommend to use so we should support this). Since I made open async I made async other methods so that their functionality is triggered after open.
```
ln1: var ref  = cordova.InAppBrowser.open()
ln2: ref.addEventListener('startload', function() {..})
ln3: ref.close()
```

PS. Thank you for the setImmediate tip - will update the code

PR #149, opened 2016-02-08
CB-10538 cordova-plugin-inappbrowser timeout issue

sgrebnov:
Switched to Apache dist as test page and increased timeout just in case (looks like there are some scripts on Apache page preventing page to be fully loaded in time in some cases).

@alsorokin PTAL as you have stable repro.


alsorokin:
Tests are green, :+1:

PR #158, opened 2016-03-24
CB-10944 : NoSuchMethodError in InAppBrowser plugin

akofman:
Sorry for that guys.


infil00p:
It's fine.  I'll do a check on Android 4.4 today to make sure this is resolved.  I missed it on this end while doing a quick test here.

PR #227, opened 2017-07-19
CB-12969 android: Added support for data URIs

matrosov-nikita:
### Platforms affected
Android

### What does this PR do?
Adds support for data Uri (e.g. `'data:text/html, <html><body><h1>Hello</h1></body></html>'`)

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,3 +707,18 @@ iab.open('http://url-that-fails-whitelist.com', 'random_string'); // loads in th
 iab.open('http://url-that-fails-whitelist.com', 'random_string', 'location=no'); // loads in the InAppBrowser, no location bar
 
 ```
+
+### Data Urls (format: data:[<mediatype>][;base64],<data>)
+
+Supported by android, windows and browser platforms 
+```
+var iab = cordova.InAppbrowser;
+iab.open('data:text/html,%3Ch1%3EHello%2C%20World!%3C%2Fh1%3E');                  // loads in the Cordova WebView (android, browser) 
+iab.open('data:text/html,%3Ch1%3EHello%2C%20World!%3C%2Fh1%3E', '_self');         // loads in the Cordova WebView (android, browser) 
+iab.open('data:text/html,%3Ch1%3EHello%2C%20World!%3C%2Fh1%3E', '_system');       // loads in the system browser
+iab.open('data:text/html,%3Ch1%3EHello%2C%20World!%3C%2Fh1%3E', '_blank');        // loads in the InAppBrowser (android, browser)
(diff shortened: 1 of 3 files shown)


filmaj:
Is this supported on any platforms other than iOS?

matrosov-nikita:
Seems that it's supported on browser platform only (for now I've verified for android, windows, ios, browser platforms). I've updated PR to add support for windows platform.
@filmaj, should I check all platforms from this list: https://github.com/apache/cordova-plugin-inappbrowser#supported-platforms?

filmaj:
The reason I ask is I am wondering if / how the documentation or the API reference for this plugin would change as a result.

And I wouldn't test against all those platforms, I personally test against Android, iOS, Windows and Browser.

matrosov-nikita:
@filmaj, I would expand this section (https://github.com/apache/cordova-plugin-inappbrowser#more-usage-info) by adding samples with Data URIs.

But, for now, I found that this patch doesn't work on windows with data URI in case of using `_blank` target. Not sure, whether it supposed to work.

filmaj:
Ideally this kind of stuff lands across all major platforms in Cordova.

Generally, if something is not doable due to platform-specific restrictions, we, at the minimum, document it (via "Quirks" in our docs).

I think it's a judgment call by you. If it's not possible to do on Windows, or the implementation is very hard and would take long, then perhaps it is worth skipping one platform implementation.

matrosov-nikita:
@filmaj, I've added a section in docs for Data URIs, could you please take a look?

filmaj:
Hey, i'm on vacation for the next couple of weeks (getting married!). CAN
you get someone else to review? Perhaps @stevesgill

On Aug 15, 2017 11:50 AM, "Nikita Matrosov" <notifications@github.com>
wrote:

@filmaj <https://github.com/filmaj>, I've added a section in docs for Data
URIs, could you please take a look?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/apache/cordova-plugin-inappbrowser/pull/227#issuecomment-322427086>,
or mute the thread

matrosov-nikita:
@filmaj, sure, my congratulations :)
@stevengill, could you please take a look?

infil00p:
Why is there a conflict in windows/InAppBrowserProxy.js? That needs to be fixed before getting merged in.

infil00p:
It looks like the PR got borked, and we can't accept it as is due to the conflict.,  Closing.

jjtfsalgado:
I'm still getting this issue. Is there any workaround?

PR #239, opened 2017-09-05
Enabling Appcache for the android webview

walisc:
Enabling App Caching for the android webview instance. Disabled by default.


janpio:
What exactly is AppCaching in this context?

andreszs:
Is this the same cache that caches all images as they are requested by the app? (not the images from an eventual webview control, I mean every single image downloaded and shown dynamically)

The current app cache has **no limit** and grows until the internal device memory has been depleted, which is renders the entire device unusable. Putting a limit to such cache would be a great idea.

walisc:
Hi @janpio and @janpio Apologies for this pull request. Was intended for my company's branch. We need to enable AppCaching for the inappbrowser as we serve our app remotely (for security reasons) yet we still want it be available offline . @janpio re:- limit, yes I agree with you. the second commit removed this limit.

janpio:
No need to apologize, I was just curious what this does and if it would make sense to actually have this in the plugin for all users, maybe behind a configuration or option.

PR #243, opened 2017-09-21
CB-5758: (android/ios) support trusting self-signed certificates

brad:
### Platforms affected

Android,iOS

### What does this PR do?

Gives the ability to trust self-signed certificates

Used in a production app for 2 years

### Checklist
- [X] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [X] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -25,7 +25,9 @@ import android.provider.Browser;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
+import android.net.http.SslError;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Bundle;
 import android.text.InputType;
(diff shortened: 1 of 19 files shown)


infil00p:
The SSL code in Cordova proper pre-dates the whitelist, so all it does is ignore the error and proceed with the SSL without validating the certificate.  I'm pretty sure we actually don't want this behaviour going forward, and if we allow any self-signed certificate trusting in the future, it will be with a flag as described in CB-12232

src/android/InAppBrowser.java
```diff
@@ -879,6 +910,48 @@ private void sendUpdate(JSONObject obj, boolean keepCallback, PluginResult.Statu
         }
     }
 
+    private void addTrustedCA() throws IOException, NoSuchAlgorithmException, CertificateException, KeyManagementException, KeyStoreException {
+        Log.d(LOG_TAG, "Adding trusted certificate if it exists");
```
infil00p: We use our own LOG, not the Android log.  This is causing compile errors in Travis CI.

infil00p:
Going to close this PR, since it needs a bit of a rethink.

PR #201, opened 2016-12-12
(Android) Implement Download Files from URL on Android WebView.

Gorebar:
### Platforms affected

Android

### What does this PR do?

Allow Download Files from URL on Android WebView

### What testing has been done on this change?

Download Files from URL.

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

I needed download files from URL on Android WebView but doesn't works, then... I added `inAppWebView.setDownloadListener` method to implement that feature. Testes on Android 4.4 and above and works fine. =)

PR #221, opened 2017-05-19
CB-12834: (android) Fix bug when opening a download link

MeirBon:
### Platforms affected
Android

### What does this PR do?
It adds a dialog to request permissions when a file is downloaded in the inappbrowser. I have committed the changes using the given format, original changes are from [this pull request](https://github.com/apache/cordova-plugin-inappbrowser/pull/201)

### What testing has been done on this change?
Multiple Android devices have been tested. Different download links, cancelling the download then re-downloading.

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #244, opened 2017-09-22
CB-12834: (android) Fix bug when opening a download link

infil00p:
### Platforms affected
Android

### What does this PR do?
It adds a dialog to request permissions when a file is downloaded in the inappbrowser. I have committed the changes using the given format, original changes are from #201 

Multiple Android devices have been tested. Different download links, cancelling the download then re-downloading.

### Checklist
- [x ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -44,10 +44,14 @@
                 <param name="android-package" value="org.apache.cordova.inappbrowser.InAppBrowser"/>
             </feature>
         </config-file>
+        <config-file target="AndroidManifest.xml" parent="/manifest">
+            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+        </config-file>
 
         <source-file src="src/android/InAppBrowser.java" target-dir="src/org/apache/cordova/inappbrowser" />
         <source-file src="src/android/InAppBrowserDialog.java" target-dir="src/org/apache/cordova/inappbrowser" />
         <source-file src="src/android/InAppChromeClient.java" target-dir="src/org/apache/cordova/inappbrowser" />
+        <source-file src="src/android/InAppBrowserDownloads.java" target-dir="src/org/apache/cordova/inappbrowser" />
 
         <!--  drawable src/android/resources -->
         <resource-file src="src/android/res/drawable-hdpi/ic_action_next_item.png" target="res/drawable-hdpi/ic_action_next_item.png" />
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -93,6 +93,7 @@ public class InAppBrowser extends CordovaPlugin {
     private static final String USER_WIDE_VIEW_PORT = "useWideViewPort";
 
     private InAppBrowserDialog dialog;
+    private InAppBrowserDownloads downloads;
     private WebView inAppWebView;
     private EditText edittext;
     private CallbackContext callbackContext;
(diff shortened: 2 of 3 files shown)


MeirBon:
@infil00p Replying on what you said in #221. I have thought about that too. It is a difficult decision, but not implementing anything results in the app crashing. If implementing downloads is not an option (due to the permissions needed during compile time) we should atleast implement a catch before the app crashes. Another option might be to store the download in a tmp/cached location and open it, I might look into this as this shouldn’t require any additional permissions (read about it, but haven’t done it myself yet).

RafaelKr:
Thanks for reopening this. Could you maybe update this PR with my code from here?
https://github.com/apache/cordova-plugin-inappbrowser/pull/221#issuecomment-320635746

This passes the cookies from the WebView to the DownloadManager, so you're able to download files from locations where you need to be signed in.

infil00p:
@RafaelKr Can you send me your branch so I can pull that in? That'd be faster than having to merge that by hand.

janpio:
Hey @infil00p , there seems to be a merge conflict now. Could you maybe fix this? Thanks.

janpio:
Fixed the merged conflict, good for review now.

RafaelKr:
@infil00p Hey, janpio just brought this PR to my attention again.
A long time has passed since my last response. I just had the time to create a PR onto your branch. https://github.com/infil00p/cordova-plugin-inappbrowser/pull/1
Maybe you can merge this in? Thanks.

janpio:
@RafaelKr Oh wow, I started something here :) Maybe you can also create a PR to this repo directly, or does this only work based on this PR?

RafaelKr:
@janpio I'm changing the `InAppBrowserDownloads.java` file which is newly created in this PR. So AFAIK I can only do this on top of this PR, no?
If there's another way, just let me know :)

If this PR is merged I can do a PR directly on this Repo.

janpio:
Sounds good.

You understand the problem, if I wanted to confirm and test this I would probably best create an app with the `android.permission.WRITE_EXTERNAL_STORAGE` permission and download a file via IAP, then one without the permission and download again. Without this PR it should crash, with this PR it should ask for permission and download the file successfully. Correct?

RafaelKr:
The `WRITE_EXTERNAL_STORAGE` permission is required to write files outside of application-specific directories. `Environment.DIRECTORY_DOWNLOADS` in this case.
https://developer.android.com/reference/android/Manifest.permission.html#WRITE_EXTERNAL_STORAGE

Since API Level 23 (Android 6.0, Marshmallow) the user explictly has to grant that permission.
This PR requests access from the user and also delegates Downloads to the native Android DownloadManager.

So yes, Downloads without this PR either just don't work or will crash the App. I don't know how this is handled by Android.
___

Will you merge this PR then?

janpio:
I'll try to recreate and confirm it, then possibly merge - yes.
Maybe we should also add tests to confirm that downloads work, but I will look into that later.

janpio:
This works like a charm (tested on a Nexus 5 running on Android 6.0.1).

- With the current plugin version, just nothing happens when you click a link to download a file. 
- With the PR branch you get the "Allow ... to access photos, media, and files on your device? [Deny] [Allow]" permission popup and a toast "Download 'filename'", afterwards the file can be found in "Downloads". (If you deny the permission you get an appropriate error message)

I think this is actually a new feature, as it adds new behavior for users on Android >=6. I changed the PR title to reflect that.

LGTM. Merge?

PR #246, opened 2017-10-09
Cb 13409 - Add more configurable options for toolbar in android and ios

steinaragustli:
### Platforms affected
Android, iOS

### What does this PR do?
See https://issues.apache.org/jira/browse/CB-13409

### What testing has been done on this change?
I have tested the combinations possible with the new available configurations and have found them working fine, see in screenshots in the link on jira. These changes shouldn't break anything since they only allow new options.

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [x] Added automated test coverage as appropriate for this change.

PR #247, opened 2017-10-26
CB-13498 added closebutton option for ios

jaapjanfrans:
### Platforms affected
ios

### What does this PR do?
Adds the option on ios to enable or disable the close button in the toolbar. I needed this for an app that should not display a close button but should show the toolbar because of the previous/next navigational buttons.

Run automated tests as outlined here for platform ios

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ instance, or the system browser.
     - __hidden__: set to `yes` to create the browser and load the page, but not show it. The loadstop event fires when loading is complete. Omit or set to `no` (default) to have the browser open and load normally.
     - __clearcache__: set to `yes` to have the browser's cookie cache cleared before the new window is opened
     - __clearsessioncache__: set to `yes` to have the session cookie cache cleared before the new window is opened    
+    - __closebutton__: set to `yes` or `no` to turn the close button on or off for the InAppBrowser's toolbar (defaults to `yes`)
     - __closebuttoncaption__: set to a string to use as the __Done__ button's caption. Note that you need to localize this value yourself.
     - __disallowoverscroll__: Set to `yes` or `no` (default is `no`). Turns on/off the UIWebViewBounce property.
     - __toolbar__:  set to `yes` or `no` to turn the toolbar on or off for the InAppBrowser (defaults to `yes`)
--- a/src/ios/CDVInAppBrowser.h
+++ b/src/ios/CDVInAppBrowser.h
@@ -49,6 +49,7 @@
 @property (nonatomic, assign) BOOL location;
 @property (nonatomic, assign) BOOL toolbar;
 @property (nonatomic, copy) NSString* closebuttoncaption;
+@property (nonatomic, assign) BOOL closebutton;
 @property (nonatomic, copy) NSString* toolbarposition;
 @property (nonatomic, assign) BOOL clearcache;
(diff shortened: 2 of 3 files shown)


jaapjanfrans:
so... can this request be considered by a project maintainer? 
If any clarification is needed or if questions arise, please let me know :-)

janpio:
My iOS coding skills are too bad to review this myself, but have a look at what I posted at this other PR with the same problem: https://github.com/apache/cordova-plugin-inappbrowser/pull/246#issuecomment-346310087

surajpindoria:
The code for this looks fine and it works as described. But I am struggling to really find a reason for adding this feature.

- Why would a developer want to have an inappbrowser window with no close button?
- As a user of the app, how would you close the inappbrowser?
- Why would this be an iOS specific feature only?

To me this seems like a very specific use case that doesn't warrant adding it back to the main plugin repo.

jaapjanfrans:
Hey thanks for chiming in!

**Why would a developer want to have an inappbrowser window with no close button?**
Simple, because the inappbrowser should not be closed in the app I'm building
**As a user of the app, how would you close the inappbrowser?**
The user is not supposed to close it :-) The app is basically wrapping a web app using the inappbrowser. It runs it's own javascriptcontext in the main webview of the app. The main webview handles push notifications, log in functionallity, sharing of content to the application from other apps on the device, the inappbrowser show the web app. The user does al his/her work in the web app.
**Why would this be an iOS specific feature only?**
On android, we just do not show the toolbar at all so the problem does not exist there. But on ios, we want to retain the forward/backwards control of the toolbar because ios has no backbutton in it's ui. So we want to have a toolbar with just the navigational controls. current situation on ios is that you get both, the nav controls and the close button.  

**To me this seems like a very specific use case that doesn't warrant adding it back to the main plugin repo.**
Of course it's a specific use case, i won't argue with that. But why not offer the option when it's so easy to implement. Why limit ourselves to an 'all or nothing' toolbar :-)

janpio:
Not OP, but my idea I got when seeing this:

* OAuth signup + login flow
* App catches "Cancel" button or redirection to "success" page
* (Probably iOS only app, waiting for implementation for Android by someone else)

surajpindoria:
@jaapjanfrans,

I had a feeling that this is what you were trying to accomplish by removing the close button. Another question, have you tried to release this app to the store before? Typically Apple will not approve apps that are simply a wrapper for a website. They will also check to see how your app functions in an offline mode, which could be another reason it would get rejected. If you have a chance please check out this detailed [post](https://forums.adobe.com/message/9741806#9741806) by another Cordova contributor with regards to this situation.

With that said, I still stand by my original opinion that this should not be merged in. I agree with you that this is a trivial change, but it would be encouraging behavior that would likely result in peoples apps being rejected by Apple.

jaapjanfrans:
I understand your concern @surajpindoria 

I think we're save with our app because of the way we add some functionality that is not available in the web variant, but i share your concern.

I think it's a bit far fetched to say that this PR will lead people to the dark side of empty wrapper apps, but i also think the're more important things in the world than this little feature. We spent more time talking about it now than it takes a dev to add the feature to his own codebase if he has the need for this ;-)

So Let's close this one. Thanks for your input!

PR #260, opened 2018-02-22
Fix nav buttons on iOS and allow custom color

bentleyo:
The navigation arrows stopped working on iOS.

### Platforms affected
iOS

### What does this PR do?
Show the navigation buttons if the user doesn't have `hidenavigationbuttons` set and allow customisation of the button color.

Tested with multiple builds from my local machine.


infil00p:
Is there a JIRA issue open for this issue?  That would help with tracking this issue in the next release.  If you could open one, or find the appropriate JIRA issue, that would be helpful.  Thanks!

erl4nc:
Could this please be expedited?   This bug prevents nav buttons from showing in iOS and there is no workaround.

infil00p:
@erl4nc If you want to create a JIRA issue for it, we can move this forward.

RasimKanca:
@infil00p I have added an [issue for the navigation button issue](https://issues.apache.org/jira/browse/CB-13984).

PR #263, opened 2018-04-05
CB-14013: (android) Change the InAppBrowser to allow custom schemes for oAuth

SailingSteve:
### Platforms affected
Android

### What does this PR do?
Allows oAuth redirects that return to Android with custom schemes like "twitter://code=nnn" to be
acted on by the InAppBrowser.  This is needed for oAuth 2.0 flow.

This PR builds on  apache/cordova-plugin-inappbrowser/pull/99 and apache/cordova-plugin-inappbrowser/pull/261 and also builds on a good suggestion from NGumby and comments about viability from infil00p

Here is my actual working JavaScript app code from [https://github.com/WeVote/WebApp/blob/develop/src/js/components/Twitter/TwitterSignIn.jsx](https://github.com/WeVote/WebApp/blob/develop/src/js/components/Twitter/TwitterSignIn.jsx)
```
          inAppBrowserRef.addEventListener('customscheme', function (event) {
            oAuthLog("customscheme: ", event.url);
            TwitterSignIn.handleTwitterOpenURL(event.url);

            inAppBrowserRef.close();
          });
```

You configure the allowable schemes in a comma separated list in a new preference.  This implements whitelisting for the allowed schemes, but doesn't use the Cordova Whitelist plugin -- this approach seemed much simpler and safer.

See my actual config.xml at [https://github.com/wevote/WeVoteCordova/blob/master/config.xml](https://github.com/wevote/WeVoteCordova/blob/master/config.xml)
```   
   <preference name="AllowedSchemes" value="mycoolapp,wevotetwitterscheme" />
```

It has been tested and is working in my app  [https://github.com/wevote/WeVoteCordova](https://github.com/wevote/WeVoteCordova).  I tested to make sure that single and multiple schemes in the "AllowedSchemes" preference worked as expected. 

### Checklist


SailingSteve:
PR is here https://github.com/apache/cordova-plugin-inappbrowser/pull/263

timbru31:
@SailingSteve possible to add support for staging+mycoolapp? Currently the plus sign is excluded from your regexp (`^[a-z]*://.*?$`)

wvengen:
Wow, this is **huge** as it (finally) allows 'events' from the inAppBrowser to the Cordova app. Its use goes way beyond OAuth 2. Thanks a lot!

As a sidenote: I do see a use-case for allowing `http` and `https` in `AllowedSchemes`: deciding in the Cordova app whether a URL needs to be loaded or not. This would allow for more easily deciding which URLs are opened in the app, and which URLs are opened in the system web browser.

boynet:
I started testing it and I ran into bug, the event only working on the first time so a page with this simple html loaded into inAppBrowser:
```
config:
<preference name="AllowedSchemes" value="whatsapp,twitter" />
....
html:
<html>
<body>
<a href="whatsapp://send?text=test">click to send</a>
</body>
</html>

event:
        inAppBrowser.addEventListener('customscheme', function (event) {
            console.log(event);
        });
```

the first click everything is fine, the second click no longer trigger the event

janpio:
Can you please create a new issue @boynet? This will be forgotten here, just link to the PR in your description for context. Thanks for taking the time to report this!

siauderman:
Why not just make it launch an intent to open the default system app for the custom scheme? Something like https://github.com/ljcljc/cordova-plugin-inappbrowser

PR #268, opened 2018-04-26
CB-14048: (android) allowedSchemes check empty string fix

wtrocki:
## Motivation

The new AllowSchemes introduced with inappbrowser@3.0.0 doesn't check if  AllowSchemes contains  empty string after having being loaded, respectively only if null, which could lead to error in case a custom scheme is use but not set as white listed schema 

What I mean is that, if no preference would be set in config.xml but a custom scheme would be used (my case) then the variable allowSchemes won't be null but will contains an empty string

Fix for https://issues.apache.org/jira/browse/CB-14048


jcesarmobile:
Merged.

Please @wtrocki, don't delete the PR templates and make sure you fill all the fields next time

PR #276, opened 2018-07-16
CB-14188: Add beforeload event, catching navigation before it happens

wvengen:
### Platforms affected

Android, iOS

### What does this PR do?

Implements [CB-14188](https://issues.apache.org/jira/browse/CB-14188): support for a callback that determines whether a URL to be loaded in the inAppBrowser should be loaded or not.

On iOS & Android emulators (will test on real devices soon).

### Checklist
- [ ] Added automated test coverage as appropriate for this change: I'd be happy to add it once the feature is finished otherwise.
- [x] Added documentation for this change (added checklist item)
- [ ] Research whether this is possible asynchronously
- [ ] Use a synchronous callback

### Example

```js
app = cordova.InAppBrowser.open(url, "_blank", "beforeload=yes")
app.addEventListener("beforeload", function(e, cb) {
  if (e.url.startsWith("https://example.com/")) {
    cb(e.url); // Tell inAppBrowser to open the URL indeed.
  } else {
    alert("Opened external URL: " + e.url);
  }
}, false);
```

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -110,6 +110,7 @@ public class InAppBrowser extends CordovaPlugin {
     private static final String HIDE_URL = "hideurlbar";
     private static final String FOOTER = "footer";
     private static final String FOOTER_COLOR = "footercolor";
+    private static final String BEFORELOAD = "beforeload";
 
     private static final List customizableOptions = Arrays.asList(CLOSE_BUTTON_CAPTION, TOOLBAR_COLOR, NAVIGATION_COLOR, CLOSE_BUTTON_COLOR, FOOTER_COLOR);
 
@@ -138,6 +139,7 @@ public class InAppBrowser extends CordovaPlugin {
     private boolean hideUrlBar = false;
     private boolean showFooter = false;
     private String footerColor = "";
+    private boolean useBeforeload = false;
     private String[] allowedSchemes;
 
     /**
@@ -245,6 +247,20 @@ public class InAppBrowser extends CordovaPlugin {
         }
         else if (action.equals("close")) {
             closeDialog();
+        }
+        else if (action.equals("loadAfterBeforeload")) {
+            if (!useBeforeload) {
+              LOG.e(LOG_TAG, "unexpected loadAfterBeforeload called without feature beforeload=yes");
+            }
+            final String url = args.getString(0);
+            this.cordova.getActivity().runOnUiThread(new Runnable() {
+                @SuppressLint("NewApi")
+                @Override
+                public void run() {
+                    ((InAppBrowserClient)inAppWebView.getWebViewClient()).waitForBeforeload = false;
+                    inAppWebView.loadUrl(url);
+                }
+            });
         }
         else if (action.equals("injectScriptCode")) {
             String jsWrapper = null;
@@ -674,6 +690,10 @@ public class InAppBrowser extends CordovaPlugin {
             if (footerColorSet != null) {
                 footerColor = footerColorSet;
             }
+            String beforeload = features.get(BEFORELOAD);
+            if (beforeload != null) {
+                useBeforeload = beforeload.equals("yes") ? true : false;
+            }
         }
 
         final CordovaWebView thatWebView = this.webView;
@@ -924,7 +944,7 @@ public class InAppBrowser extends CordovaPlugin {
                     }
 
                 });
-                WebViewClient client = new InAppBrowserClient(thatWebView, edittext);
+                WebViewClient client = new InAppBrowserClient(thatWebView, edittext, useBeforeload);
                 inAppWebView.setWebViewClient(client);
                 WebSettings settings = inAppWebView.getSettings();
                 settings.setJavaScriptEnabled(true);
@@ -1085,6 +1105,8 @@ public class InAppBrowser extends CordovaPlugin {
     public class InAppBrowserClient extends WebViewClient {
         EditText edittext;
         CordovaWebView webView;
+        boolean useBeforeload;
+        boolean waitForBeforeload;
 
         /**
          * Constructor.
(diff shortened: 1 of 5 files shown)


brodybits:
Looks nice in general. A couple major comments on my part:

The idea of passing an asynchronous callback to an event listener seems really strange to me. While I can definitely understand the logic behind making it asynchronous between native and JavaScript I don't think this really justifies introducing what looks like new paradigm in event listeners. I can think of the following alternative approaches:
- event listener *synchronously* returns non-null value to indicate that application does not want to load the URL, in a similar fashion to a standard `beforeload` event handler ref: [[1]](https://eloquentjavascript.net/15_event.html#p_nu8/BUQa7r) (preferred)
- event listener throws an exception to override default behavior, in a similar fashion to `backbutton` event on Windows ref: [[2]](https://cordova.apache.org/docs/en/latest/cordova/events/events.html#windows-quirks)
- new asynchronous callback mechanism *outside* the event handler mechanism that the application can use to filter URL load requests

The code also seems to assume that there would never be multiple outstanding load requests, which I would not agree with. What if a user clicks on a couple links in succession, or the webpage JavaScript issues multiple HTTP(S) requests?

[1] <https://eloquentjavascript.net/15_event.html#p_nu8/BUQa7r>
[2] <https://cordova.apache.org/docs/en/latest/cordova/events/events.html#windows-quirks>

wvengen:
Thanks a lot for your review, @brodybits. I agree that the current approach is not ideal, and have struggled with how to add the callback. Indeed, a synchronous callback would be the best option, that would first need some research to whether Android's and iOS's WebView implementation allow this to take a little time (because it involves a Javascript invocation).

If this is not possible (or if this is ported to a platform requiring near-immediate return from its do-I-need-to-load-this-URL-function), for Cordova users it can still be a synchronous function, with the help of a workaround similar to what happens now, but without the concurrency issue you mentioned (I have an idea to do this using a custom URL scheme).

Interesting you mention the `beforeunload` event, that would be familiar to many. In our case, the confirm dialog would not be shown. It also makes sense to skip this on first page load, as that is always controlled by the inAppBrowser-invoking code.

I'll delve into this. Since I'll be on holiday for about two weeks, it may take a little time.

wvengen:
Reading up on HTML's `beforeunload` event, I do see `preventDefault()`, e.g. on [MDN](https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload), [HTML 5.0](https://www.w3.org/TR/html50/browsers.html#unloading-documents) (suggesting `e.returnValue`) and [WHATWG](https://html.spec.whatwg.org/#prompt-to-unload-a-document) (preferring `e.preventDefault()`).

I would suggest to adopt `e.returnValue` and `e.preventDefault()` in InAppBrowser's new `beforeunload` event. This is in-line with current standards, allows using the existing event handling infrastructure, and by updating the event this will also work with multiple event handlers. It does require some changes to the event object passed on to the handlers (so that returnValue can be reported back, and `e.preventDefault()` is available).

wvengen:
I'd also suggest to explicitely enable the event (with an option like `beforeunload=yes` when opening the InAppBrowser), so that we're sure there is no performance penalty (if any). We can always consider removing the option in the future, if it doesn't turn out to be an issue in practice.

janpio:
Hey, I just fixed the problem that caused Android tests to fail in `master`. Could you rebase this PR please? This should get rid of the Android failures and possibly fix all test failures for this PR.

wvengen:
@janpio still failing on Android 7 :(

janpio:
Yep, you are unfortunately hitting this: https://github.com/apache/cordova-plugin-inappbrowser/issues/307

janpio:
And passing :) (context: https://github.com/apache/cordova-plugin-inappbrowser/issues/307#issuecomment-426404322)

brodybits:
Looks OK to me, merging

janpio:
This is already merged now, but can this be added to the tests and checked automatically that it works?

wvengen:
Oh, wait, wasn't expecting this to be merged already ... I was still coming back to processing the feedback of @brodybits on the callback mechanism. I suppose these points are still valid ... ?

I'm also ok with having this PR included right now (so the feature can be used), and changing the API later (in a next major release), if somebody has a better way of implementing it. Or mark the API as unstable, so it can be changed in a point release. What shall we do?

src/android/InAppBrowser.java
```diff
@@ -1085,16 +1105,20 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
...
          * Constructor.
          *
          * @param webView
          * @param mEditText
          */
-        public InAppBrowserClient(CordovaWebView webView, EditText mEditText) {
+        public InAppBrowserClient(CordovaWebView webView, EditText mEditText, boolean useBeforeload) {
             this.webView = webView;
             this.edittext = mEditText;
+            this.useBeforeload = useBeforeload;
```
purplecabbage: Why 2 vars that will always store the same value?

brodybits:
I was hoping to unblock forward progress, looks like it wasn't so good. I cannot promise when I will get a chance to follow up again. Sorry for the troubles.

@wvengen I do stand by the feedback I gave before, would like to treat them as minor points in case anyone has a chance to do it better someday.

purplecabbage:
Not a blocker, just a little nit. 
Keep moving forwards.

wvengen:
Ah, that clears it up. And thanks for merging: with these minor points I think it indeed makes sense to include it already, and if interest in this feature would increase, it can be improved.
I'll open an issue to track this.

Thanks again for your support, @brodybits!

src/android/InAppBrowser.java
wvengen: (this is merged already, but perhaps my reply will serve as a reference)
`waitForBeforeload` is the current state, whether we need to wait a call back from JS to really load next time it is encountered. `useBeforeload` indicates whether the `beforeload` event is used at all (see the end of `shouldOverrideUrlLoading`).

jezmck:
Sorry for the simple question, is this in the current release (v3.0.0)?
I've not been able to catch this event.

fcorallini:
@jezmck same problem form me, on ubuntu 16.04

PR #278, opened 2018-08-10
CB-14234: (ios) Don't call handleOpenURL for system URLs

dpolivy:
When calling `.open()` with a target of `_system`, the InAppBrowser on iOS is both launching the URL in the system browser AND also broadcasting to open the URL within the app (calling handleOpenURL). The latter behavior is problematic in many circumstances (e.g. when you want to explicitly open a link in a browser which is a universal link handled by the app).

This commit attempts to address this by checking the return value from openURL -- if it does not open the URL successfully, then (and only then) the code falls back to broadcasting the event within the app to handleOpenURL.

### Platforms affected
iOS

### What does this PR do?
Fixes CB-14234; it checks the return value of the call to `openURL`, and only broadcasts the `handleOpenURL` event within the app if `openURL` fails to open the URL in the system.

I've done manual testing on iOS

### Checklist

### DIFF
--- a/src/ios/CDVInAppBrowser.m
+++ b/src/ios/CDVInAppBrowser.m
@@ -298,8 +298,9 @@
 
 - (void)openInSystem:(NSURL*)url
 {
-    [[NSNotificationCenter defaultCenter] postNotification:[NSNotification notificationWithName:CDVPluginHandleOpenURLNotification object:url]];
-    [[UIApplication sharedApplication] openURL:url];
+    if ([[UIApplication sharedApplication] openURL:url] == NO) {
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Looks good to me, but remove the comment, if somebody want to know why it's changed they can check the git log/blame and the commit already have the issue id, so it's easy to spot

dpolivy:
@jcesarmobile First time I've been asked to _remove_ comments from code 😄 Done, and commit/PR updated. Let me know if there's anything else needed to merge this.

jcesarmobile:
Yeah, sometimes comments are good, but for so small change and being the issue and the PR linked to the commit, it's really easy to see why this was done, so no need of extra comments. Thanks for the PR!

PR #291, opened 2018-09-04
fix(android): fix crash for target '_system'

Ramachandrajoshi:
app was crashing when the opening the links on system browser.(target is '_system').
this issue is fixed.

### Platforms affected
Android

### What does this PR do?
just added the condition to check the dialog is null or not

manual.

### Checklist
-[*] reported on several forms, blogs and its still open bug on ionic 2 or 3 native plugin.
-[*] affected on all versions of the android platform.(custom or stock rom included).

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -279,7 +279,9 @@ public class InAppBrowser extends CordovaPlugin {
             this.cordova.getActivity().runOnUiThread(new Runnable() {
                 @Override
                 public void run() {
-                    dialog.show();
+                       if(dialog != null){
+                          dialog.show();
+                       }
                 }
             });
             PluginResult pluginResult = new PluginResult(PluginResult.Status.OK);
@@ -290,7 +292,9 @@ public class InAppBrowser extends CordovaPlugin {
             this.cordova.getActivity().runOnUiThread(new Runnable() {
                 @Override
                 public void run() {
-                    dialog.hide();
+                       if(dialog != null){
+                          dialog.hide();
+                       }
                 }
             });
             PluginResult pluginResult = new PluginResult(PluginResult.Status.OK);


janpio:
What is this `dialog` that is shown here? And why and how can it be null? How is this related to "target = '_system'"?

Ramachandrajoshi:
https://github.com/apache/cordova-plugin-inappbrowser/pull/291#issuecomment-418298554
dialog is the main window(container) for of the InAppBrowsers webview.
which opens when the target='_self' or '_blank'.
dialog will be null by default, and it will be null after iab window closed also.(look in the code).

when the target='_system' it means open the url in system browser.
so iab will make intent to launch the url action view.
but on ionic3 native plugin also implements .show() method which is causing the the action to trigger
dialog.show() method. where the dialog is null. so apps will be crashed with NEP.
```java.lang.NullPointerException: 
  at org.apache.cordova.inappbrowser.InAppBrowser$2.run (InAppBrowser.java:281)
  at android.os.Handler.handleCallback (Handler.java:751)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6776)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1518)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1408)```

wrapping the dialog.show() call with null checking will fix this problem.

**Comment by janpio** [comment]
Tests are failing unfortunately:

```
1) cordova-plugin-inappbrowser-tests.tests >> open method inappbrowser.spec.4 should support loadstart and loadstop events
  - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. in http://localhost:8000/cdvtests/jasmine-2.4.1/jasmine.js (line 1909)
Executed 5 of 6 specs (1 FAILED) (1 PENDING) in 43 secs.
```

Ramachandrajoshi:
I dont know the automated testing performed on this plugin.. i will look at it soon.

janpio:
Remaining test failure is expected.

---

> when the target='_system' it means open the url in system browser.
> so iab will make intent to launch the url action view.
> but on ionic3 native plugin also implements .show() method which is causing the the action to trigger
> dialog.show() method. where the dialog is null. so apps will be crashed with NEP.

How would I reproduce this @Ramachandrajoshi? Does it only work in an Ionic app, or can I somehow also reproduce this with a plain Cordova app?

Ramachandrajoshi:
This PR is not Valid anymore. Seems to be fine on latest versions. 
This PR is tested only on Ionic 3 apps. Not tested with Plain cordova app.

PR #293, opened 2018-09-04
GH-292 android: SSL errors handling in Android

### LINKED ISSUE #292: SSL errors handling in Android
Currently when I open a website with InAppBrowser and Android which have an invalid certificate (because of a missing chaining certificate) InAppBrowser only display a black screen and dispatch two event, a "loadstart" event and a "loadend" event.

I think that InAppBrowser should fire a "loaderror" event in order to known that something bad happened with ssl.

nicolashenry:
### Platforms affected
Android

### What does this PR do?
Fix #292

I tested on my company website which currently have a chaining certificate issue (it should be fixed soon).
Maybe it should be done on all platforms but our certificate issue only occurs on Android (iOS does not care about this problem).

### Checklist
- [x] Commit message follows the format: "CB-3232: (android) SSL errors handling in Android", where CB-xxxx is the JIRA ID & "android" is the platform affected.

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -33,6 +33,7 @@ import android.graphics.drawable.Drawable;
 import android.graphics.PorterDuff;
 import android.graphics.PorterDuffColorFilter;
 import android.graphics.Color;
+import android.net.http.SslError;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
@@ -50,6 +51,7 @@ import android.webkit.CookieManager;
 import android.webkit.CookieSyncManager;
 import android.webkit.HttpAuthHandler;
 import android.webkit.JavascriptInterface;
+import android.webkit.SslErrorHandler;
 import android.webkit.ValueCallback;
 import android.webkit.WebChromeClient;
 import android.webkit.WebResourceRequest;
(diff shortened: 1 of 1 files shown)


janpio:
Ok I see this adds some handling of the error. What happens with the message then?

nicolashenry:
Before the change, the events came like this : 
```js
{type: "loadstart", url: "https://ourdomain.com/"}
```
```js
{type: "loadstop", url: "https://ourdomain.com/"}
```

After : 
```js
{type: "loadstart", url: "https://ourdomain.com/"}
```
```js
{type: "loaderror", url: "https://ourdomain.com/", code: 0, sslerror: 3, message: "The certificate authority is not trusted"}
```
```js
{type: "loadstop", url: "https://ourdomain.com/"}
```

The ```handler.cancel();``` is just here to ensure that the connection will not be processed (the default behaviour is to cancel it anyway so the 'super' call is probably already doing it).

I am not sure about what should the ```loaderror``` should have as message, I just copied description from the constants here : https://developer.android.com/reference/android/net/http/SslError

janpio:
Ah ok, so the PR makes sure the error is not just swallowed but bubbles up to the web app and can be handled there?

nicolashenry:
Yes, exactly.

nicolashenry:
I don't know why travis build is failing, the tests seem ok when I run them myself : 

screenshot_20180917-125026_hellocordova

janpio:
Very possible this is unrelated, someone will look into this (either by fixing the flaky tests in general or by giving your PR the thumbs-up anyway) in the future. I am not sure enough to do so myself, so we'll have to wait for another maintainer.

nicolashenry:
@janpio I noticed that onReceivedError is not called on http error with Android SDK >= 23 (so loaderror is never triggered) because onReceived**Http**Error should be used instead. Should I create another issue/PR for that?

janpio:
Yes please @nicolashenry - sounds not directly related, so another PR would make it much easier to review. Thanks!

mtsibulski:
I would really like to see this merged please. I am trying to upload an app to Google and they sent me an email about a SSL Error Handler vulnerability and a deadline to fix it. Here is a help center article they provided: https://support.google.com/faqs/answer/7071387
There's a chance that Google may be warning or rejecting more apps going forward :/

breautek:
> I would really like to see this merged please. I am trying to upload an app to Google and they sent me an email about a SSL Error Handler vulnerability and a deadline to fix it. Here is a help center article they provided: https://support.google.com/faqs/answer/7071387
> There's a chance that Google may be warning or rejecting more apps going forward :/

Thanks for bringing attention to this issue. I'll review/test this tonight. 

@janpio I know it's been awhile since you approved this PR. Do you think you're approval still stands?

src/android/InAppBrowser.java
```diff
@@ -1446,6 +1448,49 @@ public void onReceivedError(WebView view, int errorCode, String description, Str
...
+                    message = "The certificate has expired";
+                    break;
+                case SslError.SSL_IDMISMATCH:
+                    message = "Hostname mismatch";
+                    break;
+                default:
+                case SslError.SSL_INVALID:
+                    message = "A generic error occurred";
+                    break;
+                case SslError.SSL_MAX_ERROR:
```
timbru31, line 1475: I wonder if we should remove that one: https://developer.android.com/reference/android/net/http/SslError#SSL_MAX_ERROR

timbru31:
We've had this bug, too, due to a mis-configured Apache webserver - so the PR definitely solves an issue. 👍

nicolashenry:
I removed the deprecated constant :+1:

src/android/InAppBrowser.java
```diff
@@ -1446,6 +1448,46 @@ public void onReceivedError(WebView view, int errorCode, String description, Str
...
         }
 
+        @Override
+        public void onReceivedSslError(WebView view, SslErrorHandler handler, SslError error) {
+            super.onReceivedSslError(view, handler, error);
+            try {
+                JSONObject obj = new JSONObject();
+                obj.put("type", LOAD_ERROR_EVENT);
+                obj.put("url", error.getUrl());
+                obj.put("code", 0);
```
timbru31, line 1458: Upon a second review I'm thinking about actually putting the enum code of the `SslError` into the `code` field? Otherwise it's pretty useless.

nicolashenry: I think "code" is supposed to be http code like 404/500 etc so I defined 0 as the code because it can't be determined here.

nicolashenry: I was maybe wrong here about the fact that this is a http code however I still think that it should not be the enum code since it could not be distinguished from other error codes.

timbru31: Indeed, it seems in other cases this is the HTTP error code.

Issue #340, opened 2018-11-02
iOS CDVWKInAppBrowser evaluateJavascript method randomly gets blocked on iOS 12


jonathanli2:
Hi, 
While testing the new ios inappbrowser plugin with wkwebview on iOS 12 device, when the javascript code calls injectScriptCode method, the application may hang. 
The reason is in CDVWKInAppBrowser.m, the completing block of evaluateJavascrip method is not called,  so the method stuck in the _**while (!finished)**_ loop

_- (NSString *)evaluateJavaScript:(NSString *)script {
    __block NSString *resultString = nil;
    __block BOOL finished = NO;
    __block NSString* _script = script;
    NSLog(@"evaluateJavaScript: %@", script);
    [self.inAppBrowserViewController.webView evaluateJavaScript:script completionHandler:^(id result, NSError *error) {
        NSLog(@"evaluatejavascript result: %@, %@", result, error);
        if (error == nil) {
            if (result != nil) {
                resultString = result;
                NSLog(@"%@", resultString);
            }
        } else {
            NSLog(@"evaluateJavaScript error : %@ : %@", error.localizedDescription, _script);
        }
        finished = YES;
    }];
    
    while (!finished)
    {
        NSLog(@"CDVWKInAppbrowser evaluateJavaScript runloop");
        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate distantFuture]];
    }
    NSLog(@"evaluateJavaScript: %@", resultString);
    return resultString;
}_

The xcode console log has the below output. After calling evaluatejavascript method, it calls runloop method few times in while block, and then just stops, the app is idle.

2018-11-02 09:44:32.646767-0400 wk[5669:2032638] evaluateJavaScript: (function(w){if(!w._cdvMessageHandler) {w._cdvMessageHandler = function(id,d){w.webkit.messageHandlers.cordova_iab.postMessage({d:d, id:id});}}})(window)
2018-11-02 09:44:39.013119-0400 wk[5669:2032638] CDVWKInAppbrowser evaluateJavaScript runloop
2018-11-02 09:44:39.013583-0400 wk[5669:2032638] CDVWKInAppbrowser evaluateJavaScript runloop
2018-11-02 09:44:39.013827-0400 wk[5669:2032638] CDVWKInAppbrowser evaluateJavaScript runloop
[...]

PR #341, opened 2018-11-02
iOS CDVWKInAppBrowser evaluateJavascript method randomly gets blocked on ios 12

jonathanli2:
### Platforms affected
iOS

### What does this PR do?
fix issue[ #340](https://github.com/apache/cordova-plugin-inappbrowser/issues/340) iOS CDVWKInAppBrowser evaluateJavascript method randomly gets blocked on iOS 12

manual testing

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -400,29 +400,17 @@ static CDVWKInAppBrowser* instance = nil;
 
 //Synchronus helper for javascript evaluation
 
-- (NSString *)evaluateJavaScript:(NSString *)script {
-    __block NSString *resultString = nil;
-    __block BOOL finished = NO;
+- (void)evaluateJavaScript:(NSString *)script {
     __block NSString* _script = script;
-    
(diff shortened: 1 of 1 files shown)


jonathanli2:
This is a blocking issue, could we review and merge the change soon?

Thanks
Jonathan

janpio:
Will this also work in older iOS versions <12?

dpa99c:
@janpio yes, [WKWebView's `evaluateJavascript`](https://developer.apple.com/documentation/webkit/wkwebview/1415017-evaluatejavascript?language=objc) and the associated [addScriptMessageHandler](https://developer.apple.com/documentation/webkit/wkusercontentcontroller/1537172-addscriptmessagehandler?language=objc) used to return the async result via the postMessage API have both been present since iOS 8.

The changes in the PR by @jonathanli2 simply remove the redundant return of the synchronous result which is never used in the WKWebView implementation (my bad for not removing it originally).

janpio:
Thanks, I don't really know enough about iOS and Obj-C to make decisions like this merge myself - but as it's "your code" I went ahead @dpa99c :)

Issue #359, opened 2018-11-29
beforeload causing POST failure


keenan35i:
Hi i really was happy about this feature as i needed it for my app, however when i enable beforeload, it fails to send post data to the server on a forum submit. The page load goes through but does not send the post data to the server

`<form accept-charset="UTF-8" action="/search_results?mobile_app=true" id="login" method="post">
`    

```
activeBrowser.addEventListener('beforeload', function(e, callBack){
        console.log('before load event triggered', e, e.url);
        //prints www.someurl.com/search_results?mobile_app=true
        // so when it redirects you dont get the post data
        callBack(e.url);
      });
```

janpio:
What platform? What OS version?

I am not sure I am following. 
- Can you explain your problem and what you are trying to do a bit better? 
- How did you open the InAppBrowser window showing that page with the form? 
- Are you trying to "hook" into the POST? 
- Without the listener everything is working, but with it the POST data gets lost?

keenan35i:
ios 12, i am trying to implement cacheing with the beforeload event, however if i include the beforeload even on a simple passthrough like the example i posted, post data within a form gets lost, the server gets the hit but without any post data. Everything works ok when i dont have the event listener or beforelaod=no.

`var options = 'beforeload=yes,usewkwebview=yes,location=no,toolbar=no,allowInlineMediaPlayback=yes,hidden=yes,disableAnimation=yes,hidespinner=yes';

cordova.InAppBrowser.open(url, '_blank', options);
`

janpio:
Could you create a super simple reproduction app that helps reproduce this issue? Start a new Cordova app with `cordova create`, add the minimal code to reproduce and understand this, put it on GitHub and post a link to the repository here. Thanks.

keenan35i:
I dont have a server i can give you access to to make the server hit, i made a slightly more complete html/js structure maybe this helps.

```
<html>
<head>
  <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0,viewport-fit=cover">
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
</head>
<body>
  <form accept-charset="UTF-8" action="/search_results?mobile_app=true" id="login" method="post">
    <input type="text" name="someuser">
    <button type="submit">Continue</button>
  </form>
  testApp.init();
</body>
</html>

```

```
var testApp = {
  init: function(){
[...]

PR #362, opened 2018-12-03
(iOS & Android) Add postMessage API support

dpa99c:
### Platforms affected
Android
iOS (both UIWebView & WKWebView implementations)

### What does this PR do?
Adds support for [postMessage API](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) enabling pages loaded into the InappBrowser to post messages back to the parent Webview of the Cordova app.

For example, sending event messages associated with UI interactions such as button clicks from the wrapped page back to the parent app Webview.

Automated tests have been extended to cover the `message` event.

### Checklist
- [x ] Commit message follows the format: "GH-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ x] Added automated test coverage as appropriate for this change.

### DIFF
--- a/src/ios/CDVUIInAppBrowser.m
+++ b/src/ios/CDVUIInAppBrowser.m
@@ -337,6 +337,16 @@ static CDVUIInAppBrowser* instance = nil;
     [self.inAppBrowserViewController navigateTo:url];
 }
 
+-(void)createIframeBridge
+{
+    // Create an iframe bridge in the new document to communicate with the CDVThemeableBrowserViewController
+    NSString* jsIframeBridge = @"var e = _cdvIframeBridge=d.getElementById('_cdvIframeBridge'); if(!_cdvIframeBridge) {e = _cdvIframeBridge = d.createElement('iframe'); e.id='_cdvIframeBridge'; e.style.display='none'; d.body.appendChild(e);}";
+    // Add the postMessage API
+    NSString* jspostMessageApi = @"window.webkit={messageHandlers:{cordova_iab:{postMessage:function(message){_cdvIframeBridge.src='gap-iab://message/'+encodeURIComponent(message);}}}}";
+    // Inject the JS to the webview
+    [self.inAppBrowserViewController.webView stringByEvaluatingJavaScriptFromString:[NSString stringWithFormat:@"(function(d){%@%@})(document)", jsIframeBridge, jspostMessageApi]];
+}
+
 // This is a helper method for the inject{Script|Style}{Code|File} API calls, which
 // provides a consistent method for injecting JavaScript code into the document.
 //
@@ -348,9 +358,6 @@ static CDVUIInAppBrowser* instance = nil;
 
 - (void)injectDeferredObject:(NSString*)source withWrapper:(NSString*)jsWrapper
 {
-    // Ensure an iframe bridge is created to communicate with the CDVUIInAppBrowserViewController
-    [self.inAppBrowserViewController.webView stringByEvaluatingJavaScriptFromString:@"(function(d){_cdvIframeBridge=d.getElementById('_cdvIframeBridge');if(!_cdvIframeBridge) {var e = _cdvIframeBridge = d.createElement('iframe');e.id='_cdvIframeBridge'; e.style.display='none';d.body.appendChild(e);}})(document)"];
-
     if (jsWrapper != nil) {
         NSData* jsonData = [NSJSONSerialization dataWithJSONObject:@[source] options:0 error:nil];
         NSString* sourceArrayString = [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];
@@ -472,6 +479,22 @@ static CDVUIInAppBrowser* instance = nil;
             }
             [self.commandDelegate sendPluginResult:pluginResult callbackId:scriptCallbackId];
             return NO;
+        }else if ([scriptCallbackId isEqualToString:@"message"] && (self.callbackId != nil)) {
+            // Send a message event
+            NSString* scriptResult = [url path];
+            if ((scriptResult != nil) && ([scriptResult length] > 1)) {
+                scriptResult = [scriptResult substringFromIndex:1];
+                NSError* __autoreleasing error = nil;
+                NSData* decodedResult = [NSJSONSerialization JSONObjectWithData:[scriptResult dataUsingEncoding:NSUTF8StringEncoding] options:kNilOptions error:&error];
+                if (error == nil) {
+                    NSMutableDictionary* dResult = [NSMutableDictionary new];
+                    [dResult setValue:@"message" forKey:@"type"];
+                    [dResult setObject:decodedResult forKey:@"data"];
+                    CDVPluginResult* pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK messageAsDictionary:dResult];
+                    [pluginResult setKeepCallback:[NSNumber numberWithBool:YES]];
+                    [self.commandDelegate sendPluginResult:pluginResult callbackId:self.callbackId];
+                }
+            }
         }
     }
 
@@ -513,6 +536,7 @@ static CDVUIInAppBrowser* instance = nil;
 
 - (void)webViewDidFinishLoad:(UIWebView*)theWebView
 {
+    [self createIframeBridge];
     if (self.callbackId != nil) {
         // TODO: It would be more useful to return the URL the page is actually on (e.g. if it's been redirected).
         NSString* url = [self.inAppBrowserViewController.currentURL absoluteString];
@@ -863,6 +887,7 @@ static CDVUIInAppBrowser* instance = nil;
 - (void)viewDidUnload
 {
     [self.webView loadHTMLString:nil baseURL:nil];
+    self.webView.delegate = nil;
     [CDVUserAgentUtil releaseLock:&_userAgentLockToken];
     [super viewDidUnload];
 }
@@ -880,6 +905,7 @@ static CDVUIInAppBrowser* instance = nil;
 {
     [CDVUserAgentUtil releaseLock:&_userAgentLockToken];
     self.currentURL = nil;
+    self.webView.delegate = nil;
 
     if ((self.navigationDelegate != nil) && [self.navigationDelegate respondsToSelector:@selector(browserExit)]) {
         [self.navigationDelegate browserExit];
--- a/src/ios/CDVUIInAppBrowser.m
+++ b/src/ios/CDVUIInAppBrowser.m
@@ -887,7 +887,6 @@ static CDVUIInAppBrowser* instance = nil;
 - (void)viewDidUnload
 {
     [self.webView loadHTMLString:nil baseURL:nil];
-    self.webView.delegate = nil;
     [CDVUserAgentUtil releaseLock:&_userAgentLockToken];
     [super viewDidUnload];
 }
@@ -905,7 +904,6 @@ static CDVUIInAppBrowser* instance = nil;
 {
     [CDVUserAgentUtil releaseLock:&_userAgentLockToken];
     self.currentURL = nil;
-    self.webView.delegate = nil;
 
     if ((self.navigationDelegate != nil) && [self.navigationDelegate respondsToSelector:@selector(browserExit)]) {
         [self.navigationDelegate browserExit];
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ The object returned from a call to `cordova.InAppBrowser.open` when the target i
   - __loaderror__: event fires when the `InAppBrowser` encounters an error when loading a URL.
   - __exit__: event fires when the `InAppBrowser` window is closed.
   - __beforeload__: event fires when the `InAppBrowser` decides whether to load an URL or not (only with option `beforeload=yes`).
+  - __message__: event fires when the `InAppBrowser` receives a message posted from the page loaded inside the `InAppBrowser` Webview.
 
 - __callback__: the function that executes when the event fires. The function is passed an `InAppBrowserEvent` object as a parameter.
 
@@ -238,6 +239,7 @@ function showHelp(url) {
 
     inAppBrowserRef.addEventListener('beforeload', beforeloadCallBack);
 
+    inAppBrowserRef.addEventListener('message', messageCallBack);
 }
 
 function loadStartCallBack() {
@@ -251,6 +253,8 @@ function loadStopCallBack() {
     if (inAppBrowserRef != undefined) {
 
         inAppBrowserRef.insertCSS({ code: "body{font-size: 25px;" });
+
+        inAppBrowserRef.executeScript({ code: "webkit.messageHandlers.cordova_iab.postMessage(JSON.stringify({my_message: 'this is the message'}));" });
 
         $('#status-message').text("");
 
@@ -300,17 +304,23 @@ function beforeloadCallback(params, callback) {
 
 }
 
+function messageCallback(params){
+    $('#status-message').text("message received: "+params.data.my_message);
+}
+
 ```
 
 ### InAppBrowserEvent Properties
 
-- __type__: the eventname, either `loadstart`, `loadstop`, `loaderror`, or `exit`. _(String)_
+- __type__: the eventname, either `loadstart`, `loadstop`, `loaderror`, `message` or `exit`. _(String)_
 
 - __url__: the URL that was loaded. _(String)_
 
 - __code__: the error code, only in the case of `loaderror`. _(Number)_
 
 - __message__: the error message, only in the case of `loaderror`. _(String)_
+
+- __data__: the message contents , only in the case of `message`. _(Object)_
 
 
 ### Supported Platforms
@@ -323,7 +333,7 @@ function beforeloadCallback(params, callback) {
 
 ### Browser Quirks
 
-`loadstart` and `loaderror` events are not being fired.
+`loadstart`, `loaderror`, `message` events are not being fired.
 
 ### Quick Example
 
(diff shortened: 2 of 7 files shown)


dpa99c:
@janpio could you do a review of this PR when you get a minute?

Travis CI tests failed for `browser-safari` platform due to `ECONNRESET` so presumably caused by connection issues and would pass if re-run.

src/ios/CDVUIInAppBrowser.m
```diff
@@ -348,9 +358,6 @@ - (void)loadAfterBeforeload:(CDVInvokedUrlCommand*)command
 
 - (void)injectDeferredObject:(NSString*)source withWrapper:(NSString*)jsWrapper
 {
-    // Ensure an iframe bridge is created to communicate with the CDVUIInAppBrowserViewController
-    [self.inAppBrowserViewController.webView stringByEvaluatingJavaScriptFromString:@"(function(d){_cdvIframeBridge=d.getElementById('_cdvIframeBridge');if(!_cdvIframeBridge) {var e = _cdvIframeBridge = d.createElement('iframe');e.id='_cdvIframeBridge'; e.style.display='none';d.body.appendChild(e);}})(document)"];
```
janpio: Is `createIframeBridge` equal to this?
Is it safe to move this to where `createIframeBridge` is called now?

janpio:
Restarted the tests that now pass.

What would be a use case for this new functionality, what problem does it solve?

src/ios/CDVUIInAppBrowser.m
dpa99c: > Is `createIframeBridge` equal to this?
Yes, this code has been factorised out into `createIframeBridge`

> Is it safe to move this to where `createIframeBridge` is called now?
In this original implementation, the iframe bridge is created "on demand", the first time that an injection method (such as `injectScriptCode`) is called. Subsequent calls find the iframe bridge already exists, so it's only created once.

In the new implementation, `createIframeBridge` is called after every successful page load (`webViewDidFinishLoad` to ensure the bridge always exists, regardless of whether an injection method has been called. This enables pages contained within the IAB Webview to make use to the bridge in order to send `message` events back to the app Webview, without relying on an injection method having been previously called in order to create the bridge.

Therefore, I think it's safe enough to do this.

src/ios/CDVUIInAppBrowser.m
```diff
@@ -863,6 +887,7 @@ - (void)viewDidLoad
 - (void)viewDidUnload
 {
     [self.webView loadHTMLString:nil baseURL:nil];
+    self.webView.delegate = nil;
```
dpa99c: Note: this code nulls the webView delegate on closing the IAB, in order to resolve an existing bug in the current UIWebView iOS implementation:

I have an app with ~10k daily iOS users and I was logging 10-15 crashes per day in Crashlytics due to this. The Crashlytics crash dump is as follows:

	#0 Crashed: WebThread EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000172df1210
	0	libobjc.A.dylib	objc_msgSend + 16
	1	UIKitCore	-[UIWebView webView:resource:canAuthenticateAgainstProtectionSpace:forDataSource:] + 92
	2 WebKitLegacy	CallResourceLoadDelegateReturningBoolean(bool, void (*)(), WebView*, objc_selector*, objc_object*, objc_object*, objc_object*) + 76
	3	WebCore	WebCore::ResourceLoader::canAuthenticateAgainstProtectionSpace(WebCore::ProtectionSpace const&) + 64
	4	WebCore	WebCore::ResourceLoader::canAuthenticateAgainstProtectionSpaceAsync(WebCore::ResourceHandle*, WebCore::ProtectionSpace const&, WTF::CompletionHandler<void (bool)>&&) + 32
	5	WebCore	WTF::Function<void ()>::CallableWrapper<-[WebCoreResourceHandleAsOperationQueueDelegate connection:canAuthenticateAgainstProtectionSpace:]::$_3>::call() + 136
	6	JavaScriptCore	WTF::dispatchFunctionsFromMainThread() + 308
	7	Foundation	__NSThreadPerformPerform + 336
	8	CoreFoundation	__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
	9	CoreFoundation	__CFRunLoopDoSource0 + 88
	10	CoreFoundation	__CFRunLoopDoSources0 + 176
	11	CoreFoundation	__CFRunLoopRun + 1040
	12	CoreFoundation	CFRunLoopRunSpecific + 436
	13	WebCore	RunWebThread(void*) + 592
	14	libsystem_pthread.dylib	_pthread_body + 128
	15	libsystem_pthread.dylib	_pthread_start + 48
	16 libsystem_pthread.dylib	thread_start + 4


The current IAB implementation sets a UIWebView delegate but never unsets it, which leads to the webview thread referencing a non-existent UIWebView instance after the IAB window is closed, due to the still-present delegate reference.

This corresponds to the [Apple documentation for UIWebView delegate](https://developer.apple.com/documentation/uikit/uiwebview/1617937-delegate?language=objc):

 > Before releasing an instance of `UIWebView` for which you have set a delegate, you must first set its delegate property to `nil`. This can be done, for example, in your dealloc method.

This code solves the issue by nilling the delegate on the webview when closing the IAB.
I have released an app update containing my own fork of this plugin which contains this fix and I have now seen these crashes stop being logged in Crashlytics in the new version.

If you'd rather this fix but pulled out and submitted as a separate PR, let me know.

janpio: Would be awesome if this was a separate PR - I think this could be related to some memory leak reports we were also getting?

dpa99c:
> What would be a use case for this new functionality, what problem does it solve?

Currently we can inject code into the page inside the IAB Webview, using `injectScriptCode` for example, and return a result.
However this only works if the result of the injected code can be obtained synchronously (i.e. `return foo`).
If the result of the injected code is generated asynchronously, we cannot currently easily return the result without a messy workaround, such as storing the result in a temporary local variable and using another `injectScriptCode` call to retrieve it.

This PR enables asynchronous communication from page in the IAB Webview back to the parent Cordova app Webview via the `message` event.
The contained pages posts a message using the postMessage API and the Cordova app listens for and receives it.
This enables data to be sent asynchronously from the page inside the IAB Webview back to the Cordova app.
This is particularly useful when the asynchrous result is not directly created by the injected code: for example, if the injected code adds an event listener to a UI element in the IAB page and it is user-interaction with that UI element that generates the result (see below).

<hr/>

A real-world use case in which I have used this functionality:

I have a Cordova app which wraps a 3rd party website in the IAB. That website consists of various promotional pages which are CMS-managed and have volatile content (marketing team is constantly changing them) so cannot be placed statically inside the Cordova app. Some of those pages contain "Buy ticket" buttons which, if clicked in a desktop browser, open a popup window containing a pre-fill form to buy certain tickets.

However when that page is loaded into the IAB, pressing that button does nothing (we can't open popup windows from the IAB).
Therefore I use `injectScriptCode` to attach `click` event listeners to those buttons which, when pressed by the user, returns the parameters which would be passed to the form in the popup window back to my Cordova app Webview via the new `message` event. 
My Cordova app is then able to listen for those `message` events and on receiving, close the IAB window and show its own native "Buy tickets" screen (complete with Apple Pay, Android Pay, etc.) with the fields pre-filled using the parameters sent back in that `message` event.

src/ios/CDVUIInAppBrowser.m
dpa99c: OK, I will create a separate issue and PR for it.

But should I also remove this code fix from this PR (since it will be added by the above PR anyway)?

janpio: If it is not too much work, you can do that.
Otherwise, we just make sure to merge your new PR first, then this will just be already present when merging this one, thus being credited as a change of the other PR.

janpio:
That does indeed sound like a _very_ useful thing to do and pretty common use case. 

You might want to create a demo app and write a blog post about it... might come in handy when we release a new version of `cordova-plugin-inappbrowser` and need something to write about and link to in the announcement post ;)

dpa99c:
> You might want to create a demo app and write a blog post about it... might come in handy when we release a new version of `cordova-plugin-inappbrowser` and need something to write about and link to in the announcement post ;)

Here's one I prepared earlier 😁 
https://github.com/dpa99c/cordova-plugin-inappbrowser-test

This tests demonstrates the new `message` event (as well as other existing IAB functionality).

src/ios/CDVUIInAppBrowser.m
dpa99c: redundant nulling statements removed in latest commit https://github.com/apache/cordova-plugin-inappbrowser/pull/362/commits/03deac68fb85a1454392d8aba4b717ff38d723f8

keenan35i:
Nice! I would definitely use this feature, would allow me to remove a lot of hacky code i wrote to communicate data back and fourth from parent to child and back.

janpio:
Can you test this branch @keenan35i and see if it works for you? Would be a great help.

dpa99c:
For reference you can install my branch using:

    cordova plugin add https://github.com/dpa99c/cordova-plugin-themeablebrowser#GH-362-postMessage

Some out-of-the-box test code you can execute in any Cordova app Webview which contains [this branch](https://github.com/dpa99c/cordova-plugin-themeablebrowser/tree/GH-362-postMessage) of the InAppBrowser plugin:

    var url = "https://braintree.github.io/popup-bridge-example/this_launches_in_popup.html";
    iab = cordova.InAppBrowser.open(url, '_blank');
    iab.addEventListener('message', function(e) {
        var msg =  JSON.stringify(e);
        iab.close();
        console.log("Message received: " +msg);
        alert(msg);
    });
    iab.addEventListener('loadstop', function (e) {
        iab.executeScript({
            code: "(function() { " +
                "var body = document.querySelector('body'); " +
                "var button = document.createElement('button'); " +
                "button.innerHTML = 'Click Me'; " +
                "body.appendChild(button); " +
                "button.addEventListener(\"click\",function(e){" +
                    "var message = {my_message: 'Hello World!'};" +
                    "webkit.messageHandlers.cordova_iab.postMessage(JSON.stringify(message));" + 
                "},false);" +
            "})(); "
        });
    });

dpa99c:
@infil00p @stevengill @agrieve @shazron @purplecabbage @jcesarmobile 
Could I bug any of you guys to review the native code changes?

keenan35i:
So i went ahead and tested it, doesnt seem to work in the console for just a basic check am i missing anything? parent is on the right child is on the left

image

dpa99c:
@keenan35i You have passed a simple string rather than stringified JSON to the `postMessage()` function: The input type for `postMessage()` must be an stringified JSON object, not a plain string.

The [spec for the `message` parameter](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Syntax) dictates that the input type must be an Object (rather than a simple String), but the reason for the Object being stringified is that in the iOS WKWebView implementation, the `postMessage()` function called by the page inside the InappBrowser is actually the native WebKit implementation and this itself expects a stringified JSON argument.
In order for the iOS UIWebView and Android implementations to conform to the same API, they have been implemented to expect the same.

This is in the [updated documentation](https://github.com/dpa99c/cordova-plugin-themeablebrowser/tree/GH-362-postMessage#inappbrowserevent-properties):

> data: the message contents , only in the case of `message`. `(Object)`


So if I call `postMessage()` with a stringified JSON Object, it works:

app
iab


But currently if I call `postMessage()` with a stringified JSON Object, it fails silently:

app_2
iab_2

In the case of Android and iOS UIWebView, we have custom implementations of `postMessage()`, so we can potentially handle this more gracefully and log an error to either the app or inappbrowser JS console if the input parameter type to `postMessage()` is incorrect.
But in the case of iOS WKWebView, because we are calling the native implementation of `postMessage()`, IIRC when I passed it an invalid argument type, it also failed silently (I will double check this and post the result).

keenan35i:
oh i see, just tested it with an object and seems to work perfect!

mattdsteele:
I have a use case that looks very similar to this. I'm trying to open an IAB to a login page hosted outside our Cordova app, and would like to close the IAB after the user successfully logs in (and has received their session cookies).

@keenan35i If this PR will take some time to implement, would you be willing to share the hacky workaround code, so I can continue developing in the interim? 😄

dpa99c:
@mattdsteele 

> If this PR will take some time to implement, would you be willing to share the hacky workaround code, so I can continue developing in the interim?

This PR is complete (including documentation and examples) and ready for integration into the main plugin codebase - it is only awaiting review and approval. You can install and use the functionality using my branch mentioned above.

purplecabbage:
Thanks @dpa99c 
What differences are there between this implementation and the postMessage available in the full browsers?
I am assuming that this does not work in cordova-browser, cordova-windows, ... If so, can you please document in the readme (as quirks) that the api/event is only available on iOS and Android targets?

@mattdsteele 
Does straight OAuth2 not work for you?
https://medium.com/@purplecabbage/oauth2-in-apache-cordova-3a3ba059b184

dpa99c:
> What differences are there between this implementation and the postMessage available in the full browsers?

The [postMessage API](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) as implemented by desktop browsers is intended for cross-frame communication, hence the API reflects this:

	targetWindow.postMessage(message, targetOrigin, [transfer]);

In the case of inappbrowser we are of course emulating this for cross-webview communication so the situation is different:

The `targetOrigin` and `transfer` parameters are there for cross-frame security hence aren't relevant.

The `message` parameter is defined by the spec as:

> Data to be sent to the other window. The data is serialized using the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/DOM/The_structured_clone_algorithm). This means you can pass a broad variety of data objects safely to the destination window without having to serialize them yourself. 

This implies that unserialized Javascript objects can be passed as the `message` and that the implementation will serialize them.
However I found that to make the 3 implementations work harmoniously, serialising the objects as a string before passing them as the `message` argument to `postMessage()` was necessary.
I have updated the documentation on my branch to make this more obvious.

Note that in the case of WKWebView, it is the only implemntation where the `postMessage()` function is native. It is implicitly available as `webkit.messageHandlers.cordova_iab.postMessage()` because `cordova_iab` is the name of the WKScriptMessageHandler registered for the WKWebView instance aand therefore the `postMessage()` function gets exposed in this namespace.
Since the Android and iOS UIWebViews don't make the postMessage API available natively, this PR creates emulations of it for those webviews under the same JS namespace as the native WKWebView one.



> I am assuming that this does not work in cordova-browser, cordova-windows, ... If so, can you please document in the readme (as quirks) that the api/event is only available on iOS and Android targets?

Docs updated to make this clearer.

dpa99c:
The Travis CI tests seem quite flaky: another couple of failures for no good reason 😞

dpa99c:
@purplecabbage I've tweaked as requested - anything else that concerns you, or do you think this is ready for merging? (@keenan35i is [chomping at the bit](https://github.com/apache/cordova-plugin-inappbrowser/issues/359#issuecomment-446388817) 😀 )

purplecabbage:
Looks good to me.

dpa99c:
@janpio: if @purplecabbage is happy, can this be merged or does it need further review?

janpio:
Thanks for pinging me ;)

yelban:
(iOS & Android) Add postMessage API support (#362)

postMessage is not working in iOS 12.0
( android is work like a charm )

test process
UIWebView APP side

> function messageCallBack(params){
>     console.log(params.data);
> }
> 
> var ref = cordova.InAppBrowser.open('http://apache.org', '_blank', 'location=yes');
> 
> ref.addEventListener('message', messageCallBack);

inAppBrowser side

> var message = 'this is the message';
> var messageObj = {my_message: message};
> var stringifiedMessageObj = JSON.stringify(messageObj);
> webkit.messageHandlers.cordova_iab.postMessage(stringifiedMessageObj);

freescout-helpdesk:
Is it already included in 3.1.0?
```
    <plugin name="cordova-plugin-inappbrowser" spec="~3.1.0" />
```

`webkit` seems to be undefined in 3.1.0 inside of the inappbrowser page:
```
alert(webkit.messageHandlers.cordova_iab.postMessage); // undefined
alert(window.webkit.messageHandlers.cordova_iab.postMessage); // undefined
alert(window['webkit'].messageHandlers['cordova_iab'].postMessage); // undefined
```

Are any other plugins required to use `webkit.messageHandlers.cordova_iab.postMessage`?

freescout-helpdesk:
Solution.

It takes time for inappbrowser to define `webkit` object. So it's impossible to use `webkit` object right away in JS, it can be used only after some time.

This way it works:
```
setTimeout(function() {
	alert(typeof(webkit));
	webkit.messageHandlers.cordova_iab.postMessage(JSON.stringify({'data': 'value'}));
}, 2000);
```

JoshuvaGeorge03:
I got an type error, when I try to do post message using 

`webkit.messageHandlers.cordova_iab.postMessage` and the error was `Type error: webkit is undefined and not an object`.  This only occurs in ios and not in android.

I was using cordova-plugin-inappbrowser v3.2.0

kindly advice, if i am doing something wrong.

JoshuvaGeorge03:
After investigating, I have found out that this issue only present in **cordova-plugin-inappbrowser v3.2.0**... And this version not rise any issue with android, but cause issues with ios as I mentioned earlier. 

So, kindly fix this issue.

NiklasMerz:
@JoshuvaGeorge03 Please open a new issue with your problem description and link it here. Reporting bugs in merged or closed PRs and issue might get them lost.

Did you use WKWebView before updating to version 3.2.0.

JoshuvaGeorge03:
@NiklasMerz Sorry, will create the new issue.

Yes, My ionic app using WKWebView engine. is it mandatory to use WKWebView Engine for version 3.2.0?

NiklasMerz:
Yes it is

JoshuvaGeorge03:
@NiklasMerz kindly take a look at issue https://github.com/apache/cordova-plugin-inappbrowser/issues/613

JoshuvaGeorge03:
> Yes it is

Could you tell me is there any other requirement? and also i have created one repo for reproducing this issue.. kindly take a look at issue, which i have created.

PR #389, opened 2018-12-31
fix: stack overflow when trying to change src of iframe

### LINKED ISSUE #388: stack overflow when trying to open an iframe
When trying to open an iframe, the code tries to get the original window.open clobber using modulemapper.getOriginalSymbol(window, 'open');
But the clobber you registered in plugin.xml is 'window.open'
This causes an infinite loop resulting in stack overflow.

regevbr:
closes #388 

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -37,7 +37,7 @@
     <platform name="android">
         <js-module src="www/inappbrowser.js" name="inappbrowser">
             <clobbers target="cordova.InAppBrowser.open" />
-            <clobbers target="window.open" />
+            <clobbers target="open" />
         </js-module>
         <config-file target="res/xml/config.xml" parent="/*">
             <feature name="InAppBrowser">
@@ -71,7 +71,7 @@
     <platform name="ios">
         <js-module src="www/inappbrowser.js" name="inappbrowser">
             <clobbers target="cordova.InAppBrowser.open" />
-            <clobbers target="window.open" />
+            <clobbers target="open" />
         </js-module>
         <config-file target="config.xml" parent="/*">
             <feature name="InAppBrowser">
@@ -109,7 +109,7 @@
     <platform name="osx">
         <js-module src="www/inappbrowser.js" name="inappbrowser">
             <clobbers target="cordova.InAppBrowser.open" />
-            <clobbers target="window.open" />
+            <clobbers target="open" />
         </js-module>
         <config-file target="config.xml" parent="/*">
             <feature name="InAppBrowser">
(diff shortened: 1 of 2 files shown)


regevbr:
Fixes #388

regevbr:
@janpio can you please review it?

janpio:
No, I don't have the time right now.

janpio:
@timbru31 We can control the commit message ourselves when merging when using Squash Merge, which should be used most of the time anyway. (I am using ["Refined Github"](https://github.com/sindresorhus/refined-github) which automatically puts the PR title into the commit message for example - so renaming the PR both cleans up the PR list and creates better commit messages)

But: The issues is missing information on how to reproduce this. I don't think this should be merged without actually seeing a reproduction and then confirming that this code change fixes the problem.

regevbr:
@janpio it happend to me and I have been using my fork (with this fix) in production ever since without a fail. @mydoal also confirmed he has that issue.

The code that caused it for me (at least I think - I didn't touch it in a long time) was the following scenario:
* My html code had an empty `iframe` element with empty `src` attribute (so it is hidden).
* When I dynamically set the `src` so it will load the contents, it happens.

regevbr:
@janpio @timbru31 actually I think the fix should be different.

Looking at https://mozilla-cordova.github.io/pages/02-plugin-development/ we can see that the actual `clobbers` target should not contain `window` and then the line I fixed should work as is.

The reason the clobber still works is the fact that `window` has a property called `window` that points to itself.
```js
window.window == window // true
```
I believe that a better fix will be to revert my change and change __all__ the clobber statements in `plugin.xml` for example
https://github.com/apache/cordova-plugin-inappbrowser/blob/84dfb09edda24c33d3a96f7b117531da8bf180c4/plugin.xml#L111
should be changed to 
```xml
            <clobbers target="open" />
```

If you agree with that, I can change the PR accordingly.

What do you think?

janpio:
https://mozilla-cordova.github.io/pages/02-plugin-development/ is not an official resource regarding Cordova plugin development, but it may very well be true. 

https://cordova.apache.org/docs/en/9.x/plugin_ref/spec.html#clobbers also confirms that `cobberls` is "Used to specify the namespace **under** `window` object where module.exports gets inserted." (my highlight) confirms that more or less.

Interesting that this never cause issues before.

(I will continue the discussion about a repro in https://github.com/apache/cordova-plugin-inappbrowser/issues/388 - we can talk about the possible fix here)

regevbr:
@janpio it works because
```js
window.window == window // true
```
I will create the reproduction repo so we can continue the discussion

regevbr:
@janpio not sure why all the builds are failing...

janpio:
Because we messed up CI over here ;) Give me a minute to fix.

janpio:
Worked.

regevbr:
@janpio any updates on the matter?

janpio:
No. The `window.open` thing is quite complicated, so someone will have to spend some serious thoughts on if we can just do it without breaking anything else (that is not covered by tests).

pradeep8093 (approved):

timbru31:
I think this PR is obsolete, because we have removed the window.open clobber. I'm closing this one and probably the linked issue, too.

PR #395, opened 2019-01-09
Android: Use evaluateJavascript execute script (not prompt)

### LINKED ISSUE #303: Android: executeScript failing for large return values (length greater than 10240)
Since Android updated to webview v 69.0.3497.91, any script that runs that returns content longer than 10240 bytes fails as the message gets truncated

Bug appears here: https://github.com/apache/cordova-plugin-inappbrowser/blob/master/src/android/InAppChromeClient.java#L114

`JSONException` occurs as message is invalid JSON (truncated string)

Bug has been lodged with Chrome but they won't fix it: 

https://bugs.chromium.org/p/chromium/issues/detail?id=884100

In the bug they report:

```
I don't think you should rely on prompt() as a means of passing data from JS to Java. There are other APIs that are properly suited to this task.
```

I am about to try a few ideas either `addJavascriptInterface` or `evaluateJavascipt` instead of `prompt` but would be open to suggestions of other interfaces

ianhowe76:
### Platforms affected

Android

### What does this PR do?

Use `evaluateJavascript` to call javascript with callback for Android instead of Prompt
Fixes #303 seen with Chrome v69.0.3497.91 when callback response is greater than 10240 bytes

Added test case where callback has response length > 10240 bytes

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -73,6 +73,7 @@ import org.apache.cordova.CordovaWebView;
 import org.apache.cordova.LOG;
 import org.apache.cordova.PluginManager;
 import org.apache.cordova.PluginResult;
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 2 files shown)


DuBistKomisch:
:+1: works for me

PR #407, opened 2019-01-22
fix custom protocol android

SalehAlhjji:
fix custom protocol android


jcesarmobile:
The plugin already have a system to allow other schemes, so adding a hardcoded "app" scheme is not useful, and also editing the plugin id is not good, closing.

Issue #417, opened 2019-02-09
inappbrowser with usewkwebview not open the links with <a href=.. target=_blank>


mosababubakr:
My app using inappbrowser to open external link and work fine, and the page with that link has more links inside, 
After apple change the webview, I try to use usewkwebview=yes  in options, but there is a bug in the inappbrowser for links that opened as new tab:
e.g.:  <a href= target="_blank"></a>

Inappbrowser will not open any links with target=_blank.

So, if the user click on that link, it will not show anything, just like there is no any links.

This happened just for links with <a href= target="_blank"></a>

So, any website has links with target blank, will not open in inappbrowser if the option usewkwebview is set to yes.
<a href= target="_blank"></a> // in html <a> for links
(this mean for browser : to open the link in new tab, but for inappbroswer, it should open in the same window.).

I try to open many websites that have some links to open in new tab (with target=_blank) via inappbrowser, but nothing happened.

Apple change the webview and inappbrowser without usewkwebview=yes will not work for ios devices, so any websites has <a href with target=_blank will not open.

Inappbrowser before i set wkwebview to yes was work fine, but after apple change the webview, some websites not work, so i set to use wkwebview to yes and work fine except any links with target = _blank.

Please solve this issue as soon as possible.

Please note that i am using the master version of cordova inappbrowser plugin to test my app.

@dpa99c 
@timbru31 
@janpio

dpa99c:
Please confirm if you are using `target=blank` or `target=_blank` since the latter is valid and the former is not - see the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)

mosababubakr:
Hello  @dpa99c 
I am using target=_blank,

I try to open many websites that using some links to open in new tab, but inappbrowser with wkwebview=yes not open those links.

Regards

dpa99c:
I could reproduce the issue on iOS/WKWebView and have implemented a solution based on https://stackoverflow.com/a/25853806/777265 as PR #418

mosababubakr:
hello @dpa99c ,

Thank you for response,

Can you please tell me when this will affect the master version ? or how can i put the plugin with this implemented solution ?

mosababubakr:
@dpa99c  @janpio  Thanks for solve the issue, but there still another issue with inappbrowser when wkwebview=yes.!
Issue details:
[...]

PR #425, opened 2019-02-11
(iOS): Fix iframes in iOS/WKWebview which were broken by #418. Fixes #424.

### LINKED ISSUE #424: Solving issue #417 generated a new issue for inappbrowser when using usewkwebview=yes
@janpio @dpa99c @agrieve @stevengill @shazron @purplecabbage @timbru31 @sgrebnov
(After the old issue https://github.com/apache/cordova-plugin-inappbrowser/issues/417 solved , there is a new issue appear in inappbrowser load errors if you open any websites. and warning messages appear from header).
When @janpio @dpa99c edited the code to resolve #417 issue, there a new issue generated.

Please check inappbrowser plugin master version and try to open any website inside inappbrowser using usewkwebview=yes, and you will see that you can't see the websites.

You can try to open: https://www.trivago.com/ in inappbrowser as an example.

Before solving issue #417 , this new issue not appear, it is just appear after issue #417 solved.

Please check

Thanks.
Regards.

dpa99c:
### Platforms affected
iOS

### Motivation and Context

This fixes iframes in WKWebView implementation for iOS which was broken by PR #418 in its attempt to fix issue #417.

#418 tried to solve the issue of page links with target=_blank not working by redirecting any request which was not targeted at the main frame back to the main frame, based on https://stackoverflow.com/a/25853806/777265.

While this fixed links where target=_blank, it also had the unintended side effect of redirecting iframe requests, causing the target URL to be loaded into the main frame.

### Description

This PR fixes that by only redirecting requests that don't have a target frame set, i.e. those with a "new window action" aka target=_blank.
This based on https://stackoverflow.com/a/25713070/777265.

### Testing

The changes have been tested with page links that contain target=_blank (e.g.  https://dpa99c.github.io/cordova-plugin-inappbrowser-test/iab_content/iab.html) and with pages that use iframes (e.g. https://www.trivago.com) to ensure both still work.

There are currently no automated tests that cover this functionality specifically.

### Checklist

- [ x] I've run the tests to see all new and existing tests pass
- [x ] I added automated test coverage as appropriate for this change
- [x ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x ] I've updated the documentation if necessary

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -573,9 +573,9 @@ static CDVWKInAppBrowser* instance = nil;
     }
     
     if(shouldStart){
-        // Fix GH-417: Handle non-default target attribute
-        // Based on https://stackoverflow.com/a/25853806/777265
-        if (!navigationAction.targetFrame.isMainFrame){
+        // Fix GH-417 & GH-424: Handle non-default target attribute
+        // Based on https://stackoverflow.com/a/25713070/777265
+        if (!navigationAction.targetFrame){
             [theWebView loadRequest:navigationAction.request];
             decisionHandler(WKNavigationActionPolicyCancel);
         }else{


purplecabbage:
lgtm!

PR #353, opened 2018-11-22
Add support for right to left direction languages

transoceanic:
### Platforms affected
Android

### What does this PR do?
fix #352 

### What testing has been done on this change?
tested on Samsung Galaxy S7 (android 7.0), Nexus 5 (v6.0), Samsung Galaxy S5 (v4.4)

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #439, opened 2019-03-06
Revert "Add support for right to left direction languages"

purplecabbage:
Reverts apache/cordova-plugin-inappbrowser#353


brodybits:
I missed the reason for reverting. Was this discussed on mailing list,
issue, or somewhere else?

I think we need a pointer to the discussion and agreement, especially since
it is on master. Did I miss anything?

purplecabbage:
Because I didn’t mean to merge it ...

PR #442, opened 2019-03-06
fix(android): Reset lefttoright if not set

### LINKED ISSUE #441: [android] lefttoright value is not reset
# Bug Report

## Problem
On Android if you spawn an InAppBrowser with lefttoright=yes and then again without lefttoright, the previous value (yes) is used. The default value is 'no' and should be reset between calls.

purplecabbage:
This fixes #441

### Checklist


timbru31:
I've stumbled across this strange behavior while reviewing #648, great that there is already a PR ;D

Issue #402, opened 2019-01-17
inAppbrowser.executeScript not working in Android


gatelli:
Hi,
I need to use executeScript like :
`inAppRef.addEventListener('loadstop', function() {
inApp.executeScript({
                    code: "document.getElementsByTagName('html')[0].innerHTML"
                  }, function(html) {
// my code using html
}
}`
This is working well with iOS but not with Android. The loadstop is triggered but not the function(html).

danova123:
Facing the same issue. It was working earlier though
```
 inAppBrowserRef = cordova.InAppBrowser.open(url, target, options);
 inAppBrowserRef.addEventListener('loadstart', loadStartCallBack);
 function loadStartCallBack() {
            inAppBrowserRef.executeScript({ code: code },executeScriptCallBack);
          }
 function executeScriptCallBack(params){
    alert("ExecuteScriptCallBack called");   ///THIS NEVER GETS CALLED

}
```

danova123:
Same issue facing here as well. It used to work earlier.

ianhowe76:
Is the `html` data in the callback large?

There is an issue on Android since Chrome v 69.0.3497.91 that truncates the data to 10240 bytes. If your callback data is greater than 10240 bytes the plugin will have an error and the callback will not be called

See #303 for details and fix (awaiting merge)

gatelli:
For my case, the callback data is >10240 bytes. I had already tried this PR by merging the 2 changed files but it's not working for me unfortunately.

danova123:
@ianhowe76 you are correct, if i just fetch one element by getElementById it works. Any idea, when your changes will be merged

ianhowe76:
@danova123 Not sure how fast PRs get merged for this project. Maybe put a comment in https://github.com/apache/cordova-plugin-inappbrowser/pull/395 saying that you tested it ok

danova123:
@ianhowe76 Unfortunately its not working for me too. 
Steps:

`cordova plugin add https://github.com/ianhowe76/cordova-plugin-inappbrowser.git`

then
`inAppRef.addEventListener('loadstop', function() { inApp.executeScript({ code: "document.getElementsByTagName('html')[0].innerHTML" }, function(html) { // my code using html } }
`

aigenseer:
@gatelli 
maybe you have chanced the file "plugins/cordova-plugin-inappbrowser/src/android/InAppBrowser.java",
[...]

PR #483, opened 2019-05-21
feat(android): Updated Support for InApp Browser in Android 8 and Above

casper123:
### Platforms affected
Android 8+

### Motivation and Context
The themeable browser is out dated as it is not being maintained by original contributors. There is a problem with inApp Browser that executeScript fails in case the returning data is greater then 10240 bytes. This issue is also opened here in inApp Browser plugin https://github.com/apache/cordova-plugin-inappbrowser/issues/402

### Description
The themeable browser is out dated as it is not being maintained by original contributors. There is a problem with inApp Browser that executeScript fails in case the returning data is greater then 10240 bytes. This issue is also opened here in inApp Browser plugin https://github.com/apache/cordova-plugin-inappbrowser/issues/402

### Checklist


janpio:
@casper123 I don't understand this Pull Request, it is clearly not mergeable. Can you explain what the idea behind it was? Did you want to create an issue?

janpio:
I am closing this for now as I see no way to fix this PR. Please let me know what to do here @casper123.

PR #496, opened 2019-06-28
(windows) blank screen issue after inappbrowser.open

Jawahars:
In windows surface simulator, setting "popup.src = strUrl;" does not seem to be triggering navigation in InAppBrowser. The screen is blank with back/forward buttons. But "popup.setAttribute('src', strUrl);" navigates.

### Platforms affected
windows

### Checklist


janpio:
What is `popup` in the code snippets you mentioned?

Jawahars:
> What is `popup` in the code snippets you mentioned?

It is webview or iframe. In my case, it was webview. The [following snippet](https://github.com/Jawahars/cordova-plugin-inappbrowser/blob/4f7cb8410bca4b05f2b212ffddabdf85065afe0d/src/windows/InAppBrowserProxy.js#L177) assigns the value to popup
`popup = document.createElement(isWebViewAvailable ? 'x-ms-webview' : 'iframe');`

janpio:
So you are opening an InAppBrowser windows from this plugin and then trying to navigate to somewhere else by changing the `src` attribute from your app that is running in Cordova? Is setting `popup.src` an documented API of this pluginß

Jawahars:
> So you are opening an InAppBrowser windows from this plugin and then trying to navigate to somewhere else by changing the `src` attribute from your app that is running in Cordova? Is setting `popup.src` an documented API of this plugin?

No, When I try to do `cordova.InAppBrowser.open(url, '_blank', option)` in windows 10 surface device, the inappbrowser is showing a blank screen with back/forward buttons. When I build with the changes in this PR, it navigates and works fine.

janpio:
Understood, finally ;)

How exactly did you create the "windows surface simulator"? What do I have to do to reproduce this?

Jawahars:
:worried: This change is not required. With further debugging, we figured out the root cause of this issue. It is not related to inappbrowser plugin, and a bug in our code. open was not working because the polyfill was unintentionally overriding the properties of x-ms-webview element. Sincere apologies, this should have been caught earlier.

PR #503, opened 2019-07-09
(android) Defensive code to prevent NULL reference exceptions for async

alexyazvinsky:
### Platforms affected
Android

### Motivation and Context

 Defensive code to prevent NULL reference exceptions for async

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -313,7 +313,9 @@ public class InAppBrowser extends CordovaPlugin {
             this.cordova.getActivity().runOnUiThread(new Runnable() {
                 @Override
                 public void run() {
-                    dialog.show();
+                    if(dialog != null){
+                        dialog.show();
+                    }
                 }
             });
             PluginResult pluginResult = new PluginResult(PluginResult.Status.OK);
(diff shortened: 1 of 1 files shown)


src/android/InAppBrowser.java
```diff
@@ -1066,12 +1070,16 @@ public void postMessage(String data) {
                 lp.height = WindowManager.LayoutParams.MATCH_PARENT;
 
                 dialog.setContentView(main);
-                dialog.show();
+                if(dialog != null){
```
NiklasMerz: If `dialog` would be null, it would crash anyway because of the lines above and below. I would put the null check around all of these lines:

```
                if(dialog != null){
                    dialog.setContentView(main);
                    dialog.show();
                   dialog.getWindow().setAttributes(lp);
                }
```

timbru31: And add a space after the `if` and before the `{` for all new guards 🙃

NiklasMerz: Good catch!

NiklasMerz:
I like to live dangerous and rebased this PR and did the suggested changes. I hope I did not mess anything up. A backup  is in my fork [here](https://github.com/apache/cordova-plugin-inappbrowser/compare/master...NiklasMerz:pr503)

That's what I did:
* rebased the PR to latest master
* did an interactive rebase to squash the two relevant commits and drop the reverted changes
* changed the author to @alexyazvinsky and pushed this as a signed commit since I did this
* Did the suggested changes seperately

I hope that makes this PR better to read and follows the Apache guidelines. We should merge this and **not squash merge** to make the authors and history clear.

src/android/InAppBrowser.java
NiklasMerz: I just pushed these changes. See my following comment

PR #528, opened 2019-08-23
Update README.md

julian-lang:
Added alternative way to get window.open function back

### Platforms affected
All

### Motivation and Context

delete window.open as stated in Documentation does not bring back original function.
Especially useful, if running in browser platform.

### Description
Added alternative Way in Documentation on how to get native window.open function back.

### Testing
Implemented it in local Project -> can use window.open natively again.

### Checklist

- [ x ] I've run the tests to see all new and existing tests pass
- [ x ] I added automated test coverage as appropriate for this change
- [ x ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ x ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ x ] I've updated the documentation if necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,11 @@ behaviour:
 
     delete window.open // Reverts the call back to its prototype's default
 
+You can also get the original (native-code) function back with:
+
+    window.open = cordova.require('cordova/modulemapper').getOriginalSymbol(window, 'window.open');
+    
+
 Although `window.open` is in the global scope, InAppBrowser is not available until after the `deviceready` event.
 
     document.addEventListener("deviceready", onDeviceReady, false);


NiklasMerz:
Thank you for this PR, someone might find it useful!

Before merging this change we may need to think again about the statemen in README:
```
For backwards compatibility, this plugin also hooks window.open. However, the plugin-installed hook of window.open can have unintended side effects (especially if this plugin is included only as a dependency of another plugin). The hook of window.open will be removed in a future major release. Until the hook is removed from the plugin, apps can manually restore the default behaviour:
```

Do we still plan to remove this backwards compatibility?

NiklasMerz:
This is no longer needed since the overwrite has been removed in #600

PR #547, opened 2019-10-06
commented closedialog at the end

mrshetab:
### Checklist

### DIFF
--- a/src/android/InAppBrowserDialog.java
+++ b/src/android/InAppBrowserDialog.java
@@ -50,7 +50,7 @@ public class InAppBrowserDialog extends Dialog {
             if (this.inAppBrowser.hardwareBack() && this.inAppBrowser.canGoBack()) {
                 this.inAppBrowser.goBack();
             }  else {
-                this.inAppBrowser.closeDialog();
+                //this.inAppBrowser.closeDialog();
             }
         }
     }


purplecabbage:
Commenting out code is never a solution.  What is this for? What issue are you trying to resolve?

mrshetab:
I used this plugin in my application, but when the user presses the hardware back button while being on the first page, this will cause the browser to get closed and not minimizing the whole app. I want the app to be minimized instead of closing the app browser.

purplecabbage:
Your change would force every app that uses this plugin to behave how you want your own app to behave.  You should look at changing this in your own app.

The android option 'hardwareback' might do what you want.

mrshetab:
Oh ok! But I have two problems.
1- hardwareback won't do the thing.
2- I use "phonegap build" to build my app and "phonegap build" would only use plugins from github and I can not download the code, implement my changes in the code and then use it as my special plugin.

purplecabbage:
PhoneGap Build CAN build any plugin published to NPM or any github repo.
Just publish your own version ( with your own name for it ... ) and you can use it with pgb.

I would however advise looking at doing some/most of this in your app.
There are ways to intercept the back button in cordova-android, and you could also detect the condition that you want to exit on ... without knowing your specific goal I cannot tell you exactly how you would accomplish it.

mrshetab:
Oh ok, Thank you very much

PR #602, opened 2020-01-05
Adding info about usesCleartextTraffic

mosabab:
The default API level in the Cordova Android platform has been upgraded. On an Android 9 device, clear text communication is now disabled by default.

### Platforms affected
Android

### Motivation and Context

After upgrading to Cordova Android 8.0, you will see the error: net::ERR_CLEARTEXT_NOT_PERMITTED errors when trying to connect to http:// targets on external websites

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ instance, or the system browser.
 
 The default API level in the Cordova Android platform has been upgraded. On an Android 9 device, clear text communication is now disabled by default.
 
-To allow clear text communication again, set the android:usesCleartextTraffic on your application tag to true in `config.xml` file:
+To allow clear text communication again, set the `android:usesCleartextTraffic` on your application tag to true in `config.xml` file:
 `<platform name="android">
   <edit-config file="app/src/main/AndroidManifest.xml" mode="merge" target="/manifest/application">
       <application android:usesCleartextTraffic="true" />
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ instance, or the system browser.
 
 The default API level in the Cordova Android platform has been upgraded. On an Android 9 device, clear text communication is now disabled by default.
 
-To allow clear text communication again, set the android:usesCleartextTraffic on your application tag to true:
+To allow clear text communication again, set the android:usesCleartextTraffic on your application tag to true in `config.xml` file:
 `<platform name="android">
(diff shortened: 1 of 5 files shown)


NiklasMerz:
Thank you for documenting this. Since this is not only applicable to InAppBrowser. This change should probably done in the general Cordova Android documentation: https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/guide/platforms/android/index.md


It may be helpful for developers with certain requirements, but users should know what this means in terms of security. Maybe we should add this to the docs with a warning from the Android documenation:

> Indicates whether the app intends to use cleartext network traffic, such as cleartext HTTP. The default value for apps that target API level 27 or lower is "true". Apps that target API level 28 or higher default to "false".
When the attribute is set to "false", platform components (for example, HTTP and FTP stacks, DownloadManager, and MediaPlayer) will refuse the app’s requests to use cleartext traffic. Third-party libraries are strongly encouraged to honor this setting as well. The key reason for avoiding cleartext traffic is the lack of confidentiality, authenticity, and protections against tampering; a network attacker can eavesdrop on transmitted data and also modify it without being detected.

mosabab:
> Thank you for documenting this. Since this is not only applicable to InAppBrowser. This change should probably done in the general Cordova Android documentation: https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/guide/platforms/android/index.md
> 
> It may be helpful for developers with certain requirements, but users should know what this means in terms of security. Maybe we should add this to the docs with a warning from the Android documenation:
> 
> > Indicates whether the app intends to use cleartext network traffic, such as cleartext HTTP. The default value for apps that target API level 27 or lower is "true". Apps that target API level 28 or higher default to "false".
> > When the attribute is set to "false", platform components (for example, HTTP and FTP stacks, DownloadManager, and MediaPlayer) will refuse the app’s requests to use cleartext traffic. Third-party libraries are strongly encouraged to honor this setting as well. The key reason for avoiding cleartext traffic is the lack of confidentiality, authenticity, and protections against tampering; a network attacker can eavesdrop on transmitted data and also modify it without being detected.

Thanks for your information, I created a pull request regarding this in this pull:
https://github.com/apache/cordova-docs/pull/1050

NiklasMerz:
Thank you for creating the new one quickly. I am closing this one as a duplicate, we will look into this.

Issue #301, opened 2018-09-22
Issue with status bar on iPhone X when toolbar is at the top (v3.0.0)


russcarver:
There is an issue with the status bar and the iPhone X as described at http://www.doroja.com/smartdic.aspx?decode=1&url=aHR0cHM6Ly9zdGFja292ZXJmbG93LmNvbS9xdWVzdGlvbnMvNDczOTk5MzgvaXNzdWUtd2l0aC1zdGF0dXMtYmFyLXVzaW5nLWNvcmRvdmEtaW5hcHBicm93c2VyLWFuZC1pb3MtMTEtYW5kLWlwaG9uZS14B64Coded!

When the inappbrowser toolbar is at the top of the screen, the status bar on the iPhone X is half white and half gray.

Changing the following line in CDVInAppBrowser.m solves the problem and makes the whole status bar gray.  Change line:
`statusBarFrame.size.height = STATUSBAR_HEIGHT;`
to
`statusBarFrame.size.height = 0;`

This does not seem to negatively affect other iPhones or iPads.

russcarver:
<img width="391" alt="screen shot 2018-09-21 at 8 12 49 pm" src="https://user-images.githubusercontent.com/4061972/45912721-e3f7fb00-bdda-11e8-9487-a83800e5940b.png">

russcarver:
<img width="407" alt="screen shot 2018-09-21 at 7 39 01 pm" src="https://user-images.githubusercontent.com/4061972/45912729-eb1f0900-bdda-11e8-810a-e6591080587c.png">

vespino:
@russcarver is this something I can fix when using Phonegap Build? Or is this an update waiting to happen?

vespino:
@russcarver were you able to solve this issue? Adding the splashscreens does not solve it and iPhone X has been around for over a year now, so you would think the guys at Phonegap would have fixed this by now.

russcarver:
@vespino I'm waiting for them to fix this. In the mean time, I've just forked it and made the changes shown above.

vespino:
@russcarver can I use your fork?

russcarver:
@vespino I don't have it anymore as I decided to go a different route, but if you fork it yourself and make the change outlined above, it will work fine. You can install plugins from local directories using the "cordova plugin add" command.

vespino:
@russcarver do you mind sharing the different route btw?

vespino:
[...]

Issue #604, opened 2020-01-06
Remove fake status bar for iOS


NiklasMerz:
There are some issues with iPhones with a notch and the status bar. PR #546 tries to fix that by removing the status bar for these devices.

This is not a proper fix, it just hides the fake status bar for iPhone X and other devices with notches.

The proper fix should make the fake statusbar the size of the notch to be consistent, or just remove the fake statusbar for all devices.

_Originally posted by @jcesarmobile in https://github.com/apache/cordova-plugin-inappbrowser/pull/546#issuecomment-542434369_

Discussion in #546 tends to the removal of the status bar in major release.

NiklasMerz:
We should try to get a consistent behaviour with Android. See #598

ashtt:
Around what time/month or in which version should we expect this fix ?

jcesarmobile:
When somebody sends a pull requests that fixes it and it gets merged and a release process is started and finish successfully.
Version >= 4.0.0

benvallack:
So how do we use this version? Also if we use _self to get the pages to load in the cordova view itself, would that improve things? Trouble is I see this issue when I try that: https://stackoverflow.com/questions/58321114/why-i-get-the-console-warning-process-kill-returned-unexpected-error-1-when

NiklasMerz:
This has nothing to do with this issue. A version which removes the statusbar does not exist, yet.

sardapv:
I have fixed that notch issue for iphoneX+ series (https://issues.apache.org/jira/browse/CB-13956)
It handles both landscape and portrait orientation change as well (right left area both). Inappbrowser scales first checking all other factors. Only problem I found is when I change to landscape mode I see status bar pane but no battery wifi or any symbol in it. Not a major issue I suppose. Currently those changes are on my local.

@jcesarmobile   @NiklasMerz 
It does similar thing that is in PR https://github.com/apache/cordova-plugin-inappbrowser/pull/546 
But does it better with orientation change even afterIAB is opened.

PR #546, opened 2019-10-05
[WIP] Add top notch detection

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

This fixes an issue with iPhone that have a notch.

Closes #301 

### Description

Detect if phone has notch and adjust statusbar.

### Testing

Using with iPhones with notch and without. Deployed to production app for some time.

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [x] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x] I've updated the documentation if necessary

PR #629, opened 2020-02-21
(ios) removing hard coded height

MoemenMostafa:
### Platforms affected
iOS

### Motivation and Context
The hard coded height is causing a problem in phones with notch when this plugin used without the statusbar plugin.
https://github.com/apache/cordova-plugin-inappbrowser/issues/301

### Checklist


NiklasMerz:
We already have similar attempts for the status bar on devices with a notch: #546

But we want to remove this "fake statusbar" altogether soon #604

NiklasMerz:
The statusbar has been removed in #656

PR #647, opened 2020-03-18
feat(android): Ask for file permission when opening a download link (CB-12834) (rebased)

yesenarman:
### Platforms affected

Android

### Motivation and Context

See #201 and #244.

### Description

I wanted to bring back attention to apache/cordova-plugin-inappbrowser#244, but noticed that the original PR contains some merge conflicts and is quite behind the master branch. I also noticed that there are a lot of formatting changes in that PR that are not quite related to the added feature, so thought that maybe creating a new PR without the formatting changes will make it easier to review and merge.

This PR also contains the cookies passing improvement by @RafaelKR.

### Testing

- Tested in the app where I needed to have this feature.
  - It asks for the file storage permissions when you try to download a file for the first time.
  - If the permissions were given, it shows a toast with a message that the file download started. You can observe the download progress through notification drawer and you can also cancel the download from there.
  - If the permissions were not given, it shows a toast with an error message.

- Followed the testing instructions from CONTRIBUTING.md
  - Auto Tests completed successfully.
  - There were some issues with Manual Tests on my device, but it seems that same issues also happen when compiling from upstream master branch.

### Checklist

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -44,10 +44,14 @@
                 <param name="android-package" value="org.apache.cordova.inappbrowser.InAppBrowser"/>
             </feature>
         </config-file>
+        <config-file target="AndroidManifest.xml" parent="/manifest">
+            <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+        </config-file>
 
         <source-file src="src/android/InAppBrowser.java" target-dir="src/org/apache/cordova/inappbrowser" />
         <source-file src="src/android/InAppBrowserDialog.java" target-dir="src/org/apache/cordova/inappbrowser" />
         <source-file src="src/android/InAppChromeClient.java" target-dir="src/org/apache/cordova/inappbrowser" />
+        <source-file src="src/android/InAppBrowserDownloads.java" target-dir="src/org/apache/cordova/inappbrowser" />
 
         <!-- drawable src/android/resources -->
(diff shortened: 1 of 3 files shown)


RafaelKr:
Awesome work!
If you're able to request a review, you should add @janpio

RafaelKr:
Maybe the Toast message text should be configurable (translatable).

yesenarman:
Thanks. I think I don't have rights to request a review.

> Maybe the Toast message text should be configurable (translatable).

That's a good point.

Should they be configurable via the options string? There are two things that concern me about passing them via options string:
1. It is also easy to forget to pass these strings every time you need to open IAB. Also the options string might become very long.
2. The success message has this format currently "Downloading File '{filename}'". So should users be able to configure/translate just the "Downloading File" part or should they be able to specify the whole string formatting, including the {filename} substitution part?

Another option would be to have these strings as Android String resources in `values/strings.xml` and then let users to provide translations by adding appropriate XML file with translated strings (e.g `values-ru/strings.xml`).

RafaelKr:
I think using Android String resources would be the best way.

Defaults can be set in the plugin.xml like this: https://github.com/ionic-team/cordova-plugin-ionic/blob/63092070ca2e45564f0c4af0f8abda2f2e728b05/plugin.xml#L63 (Using strings instead of the variables, of course)
The user can then overwrite/extend it for his languages in his config.xml by putting a `config-file` block into `<platform name="android">...</platform>`.

The approach described here could be used to substitute the filename: https://stackoverflow.com/a/3656409/4303873

timbru31:
CI failure was unrelated and it's now green.

PR #654, opened 2020-03-25
[GH-652] add check for openInSystem postNotification

### LINKED ISSUE #652: window.handleOpenURL is triggered in WKWebView for system browser calls
# Bug Report
When triggering a url to open in the `_system` browser through IAB with WKWebView window.handleOpenURL is called. If using UIWebView this does not happen. This appears to be a regression as this [issue](https://issues.apache.org/jira/browse/CB-14234)  was fixed with [this PR](https://github.com/apache/cordova-plugin-inappbrowser/pull/278) but it appears the check never made it into [CDVWKInAppBrowser.m](https://github.com/apache/cordova-plugin-inappbrowser/blob/master/src/ios/CDVWKInAppBrowser.m#L360)

## Problem
I believe there should be a check around the postNotification eg: `if ([[UIApplication sharedApplication] openURL:url] == NO)`

window.handleOpenURL is not called

window.handleOpenURL is not called

### Command or Code
I created a repo to reproduce
https://github.com/j--w/iab-handleopenurl-bug/blob/master/www/js/index.js#L30

### Version information
Cordova CLI: 7.1.0
Cordova IOS: 5.1.1
XCode 11.2.1
Cordova InAppBrowser 3.2.0
OSX

## Checklist

- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

j--w:
### Platforms affected
iOS

### Motivation and Context
fixes #652 

This fix prevents window.handleOpenURL from being triggered by calls to _system browser through InAppBrowser which was previously fixed in UIWebView version of IAB but still exists in WKWebView version

### Description
Add ` if ([[UIApplication sharedApplication] openURL:url] == NO)` check in openInSystem method which was present in UIWebView version of IAB but not WKWebView.

### Testing
Ran plugintests repo
Manually tested in my own project where I first noticed the issue
Manually tested in minimum reproduction repo and pushed [to a branch](https://github.com/j--w/iab-handleopenurl-bug/commit/e910b312cb606eee2d3c922c70bf696117cfa2f3)

### Checklist

- [X] I've run the tests to see all new and existing tests pass
- [X] I added automated test coverage as appropriate for this change
- [X] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [X] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [X] I've updated the documentation if necessary

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -357,8 +357,10 @@ static CDVWKInAppBrowser* instance = nil;
 
 - (void)openInSystem:(NSURL*)url
 {
-    [[NSNotificationCenter defaultCenter] postNotification:[NSNotification notificationWithName:CDVPluginHandleOpenURLNotification object:url]];
-    [[UIApplication sharedApplication] openURL:url];
+    if ([[UIApplication sharedApplication] openURL:url] == NO) {
+        [[NSNotificationCenter defaultCenter] postNotification:[NSNotification notificationWithName:CDVPluginHandleOpenURLNotification object:url]];
+        [[UIApplication sharedApplication] openURL:url];
+    }
 }
 
 - (void)loadAfterBeforeload:(CDVInvokedUrlCommand*)command


timbru31:
Great, thanks for your PR. Code wise it looks good. I'll test your branch later today.

Issue #409, opened 2019-01-25
Resize the IAB


piterio19:
Hello,
After some reading I understand that IAB overlays the Cordova WebView completely.
I have AdMob ads showing at the bottom of my app using AdMob Pro plugin, so when I load the IAB they get hidden.

Is there any solution to this?
Otherwise can I clone the code of IAB and reduce the height of the IAB activity by those 50px I need? Where can I do that in the code?
Thanks a lot.

jcesarmobile:
It's not possible.

You can clone and try to do it, but you'll have to figure out the where yourself.

elrafael:
Open the file `InAppBrowser.java` and search for this code

```java
WindowManager.LayoutParams lp = new WindowManager.LayoutParams();
lp.copyFrom(dialog.getWindow().getAttributes());
lp.width = WindowManager.LayoutParams.MATCH_PARENT;
lp.height = WindowManager.LayoutParams.MATCH_PARENT;
```

It's near line 958
you'll need to change the lp.height attribute to something like this
```java
// lp.height = WindowManager.LayoutParams.MATCH_PARENT;
DisplayMetrics displayMetrics = new DisplayMetrics();
cordova.getActivity().getWindowManager().getDefaultDisplay().getMetrics(displayMetrics);
int displayWidth = displayMetrics.widthPixels;
int displayHeight = displayMetrics.heightPixels - 200;
lp.height = displayHeight;
```

Import the Display Metrics on top of that file
```java
import android.util.DisplayMetrics;
```

elrafael:
If you'll need align in the top and close the inappbrowser browser outside the inappbrowser, you could do this (``InAppBrowser.java``)

Search this code
```java
dialog = new InAppBrowserDialog(cordova.getActivity(), android.R.style.Theme_NoTitleBar);
dialog.getWindow().getAttributes().windowAnimations = android.R.style.Animation_Dialog;
```

Replace with this
```java
dialog = new InAppBrowserDialog(cordova.getActivity(), android.R.style.Theme_NoTitleBar);
dialog.getWindow().getAttributes().windowAnimations = android.R.style.Animation_Dialog;
dialog.getWindow().getAttributes().gravity = Gravity.TOP;
```

And in this code block
```java
dialog.setCancelable(true);
```

Add this
```java
dialog.setCanceledOnTouchOutside(true);
```

markarupert:
[...]

PR #655, opened 2020-03-25
(ios & android) Reduce IAB Web View height from the bottom

PDLMobileApps:
### Platforms affected
iOS & Android

### Motivation and Context

This change is needed in situations where the main app has some content on the bottom of the screen that should be visible even when the IAB web view is shown.
For instance, clickable footer, ad content, etc.

This change addresses a similar issue to feature request #409, which is limited to Android only. This solution also includes iOS.

### Description

This change adds a new setting in both Android and iOS named: **bottomreduceheightby** which takes an integer as value. It defaults to 0 to maintain backward compatibility.
When a value different than 0 is specified, the IAB web view height is reduced by that amount of pixels and aligned to the top of the screen. That results in a portion of the underlying Cordova window to be shown (as much as the value assigned to the setting) so that its content is fully visible and accessible.

NOTE: if needed, a similar option can be added for the top of the IAB web view.

### Testing

Testing was done manually as I am not sure on how to automate it.

1. Created a new Cordova test app using cordova-plugin-inappbrowser.

2. Executed:
    `cordova.InAppBrowser.open(url, "_blank", "");`
    and verified the web view is occupying the entire screen (default behavior).

3. Executed:
    `cordova.InAppBrowser.open(url, "_blank", "bottomreduceheightby=0");`
     and verified the web view is occupying the entire screen (default behavior).

4. Executed:
    `cordova.InAppBrowser.open(url, "_blank", "bottomreduceheightby=60");`
    and verified the web view is occupying the entire screen minus 60 pixels at the bottom.
    Also, verified the possibility to interact with the content in the underlying Cordova window.

We repeated the above tests with both Android and iOS.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,8 @@ instance, or the system browser.
     - __mediaPlaybackRequiresUserAction__: Set to `yes` to prevent HTML5 audio or video from autoplaying (defaults to `no`).
     - __shouldPauseOnSuspend__: Set to `yes` to make InAppBrowser WebView to pause/resume with the app to stop background audio (this may be required to avoid Google Play issues like described in [CB-11013](https://issues.apache.org/jira/browse/CB-11013)).
     - __useWideViewPort__: Sets whether the WebView should enable support for the "viewport" HTML meta tag or should use a wide viewport. When the value of the setting is `no`, the layout width is always set to the width of the WebView control in device-independent (CSS) pixels. When the value is `yes` and the page contains the viewport meta tag, the value of the width specified in the tag is used. If the page does not contain the tag or does not provide a width, then a wide viewport will be used. (defaults to `yes`).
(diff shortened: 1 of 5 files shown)


rajasekarsp:
May i know when this can be merged. Its a very useful feature

MykytaPereverziev:
Hi!
Please, tell me is there any feature for changing the top border in InappBrowser? If it is already implemented how can I use it? I tried to use this branch. Footer works excellent, but designer told me that I have to move buttons block to the top.

PDLMobileApps:
> Hi!
> Please, tell me is there any feature for changing the top border in InappBrowser? If it is already implemented how can I use it? I tried to use this branch. Footer works excellent, but designer told me that I have to move buttons block to the top.

This PR only adds support to reduce IAB height from the bottom. The same idea can be extended to reduce the height from the top as well, but that's not part of this PR.

Nida013:
Hi, @NiklasMerz Can you please include this commit in the upcoming release? It would be very helpful for us if you could add this, as it is a very useful feature.

Issue #623, opened 2020-02-10
Build with Xcode11 causes status bar rotate


fcojavierMT:
# Bug Report

## Problem

Hi,
If I build my app with Xcode11, after the first InApp browser plugin call, the status bar starts to rotate even if my screen orientation is forced to portrait. If I come back and build it with Xcode10, everything is fine.

### What is expected to happen?

Status bar don't rotate, like on Xcode10.

### What does actually happen?

Status bar rotates.

## Information

Xcode: 11, plugin version: 3.2.0

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

rricamar:
Hi @fcojavierMT , I was reading your issue and I think is completely the same that one that I opened time ago: https://github.com/apache/cordova-plugin-inappbrowser/issues/564

Could you check the video that I left and confirm it?

fcojavierMT:
Hi @rricamar, yes i saw your video and is completely the same issue. Hope someone of the team could help use since April Xcode10 is going to be deprecated! :cry:

sergicollado:
I'm suffering the same unexpected behaviour :(

rricamar:
Hi @NiklasMerz , sorry in advance for mentioning you here, but I saw that you help in another issues of the repo. Do you have any idea of how solve this or some insight? Are we using the plugin in a non-desired way? Ty for any help you could provide 🙏 👍

NiklasMerz:
Please check #604 and related PRs and issues. There are plans to remove the fake statusbar in this plugin for the next version. There is not PR yet as far as I know.

rricamar:
Hi @NiklasMerz ty again for helping with some info, really appreciated, I will read both issue and PR and keep following this, ty again 😇

jcesarmobile:
How are you locking the screen rotation? Can you provide a sample app? I can't reproduce

rricamar:
Hi @jcesarmobile , ty for you reply,

On my case, the flow is the next:

1. Using https://github.com/apache/cordova-plugin-screen-orientation I block the `screen-orientation` to only portrait. It works since in our app we have a public part which you can navigate without trying to login and the status bar doesn't rotate on that part
[...]

PR #672, opened 2020-04-16
fix(ios): prevent statusbar rotation after closing InAppBrowser

jcesarmobile:
### Platforms affected
ios

### Motivation and Context
closes https://github.com/apache/cordova-plugin-inappbrowser/issues/623

### Description

according to the stackoverflow link, setting the tmpWindow to hidden should release it and the regular window would take control, but it's not being released and that causes the statusbar to rotate (because the tmpController can rotate to any direction).
Setting the tmpWindow to nil releases it and the original window takes control and fixes the issue.

### Checklist

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -324,6 +324,7 @@ static CDVWKInAppBrowser* instance = nil;
     // Set tmpWindow to hidden to make main webview responsive to touch again
     // https://stackoverflow.com/questions/4544489/how-to-remove-a-uiwindow
     self->tmpWindow.hidden = YES;
+    self->tmpWindow = nil;
 
     if (self.inAppBrowserViewController == nil) {
         NSLog(@"Tried to hide IAB after it was closed.");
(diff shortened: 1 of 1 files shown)


jcesarmobile:
not really sure what's the difference between hide and close, but if hide is supposed to not destroy the IAB, then probably shouldn't be done there as it might destroy the IAB session too, but didn't try

jeirasmus:
Nice fix, but not fully working at least with our application. If you close the inappbrowser in landscape mode and application under it returns to be in portrait mode, view is a little bit higher that. If you keep device on portrait all the time, that does not happen.

jcesarmobile:
That’s a different problem

jcesarmobile:
@timbru31, I tested and the rotation was happening when using hide, so I've set it to nil on hide too.
Setting the window to nil doesn't affect the IAB VC, if you call show later it keeps showing last page it was showing before calling hide (meaning, IAB VC doesn't get destroyed, only the tmpWindow, and the tmpWindow gets recreated on show method).

PR #685, opened 2020-04-27
chore: update install engines

jcesarmobile:
master version was bumped to 4.0.0-dev, so the package.json should have a new "5.0.0" entry or will fail to install once released.

Updated the "4.0.0" to require `cordova-ios >=6.0.0-dev` since the ios breaking changes will only work in that new version.

Also updated the engine in `plugin.xml` for also require `cordova-ios >=6.0.0-dev`.

Kept the `cordova >=3.1.0` as nothing was changed to require a newer version, but maybe we should require at least 9?


NiklasMerz:
Tests fail with:

```
  cordova-plugin-inappbrowser-tests.tests >>

    cordova.InAppBrowser

      ✗ inappbrowser.spec.1 should exist

        - Expected undefined to be defined.
```

Maybe the dependencies cannot be loaded because the -dev versions are not on NPM?

jcesarmobile:
Yeah, since this requires cordova-ios 6, test won’t work until it’s released. But better have it here so we don’t forget.

jcesarmobile:
Looks like I was wrong and cordova-ios 6 is not required for IAB to work, master is working fine on 5.1.1.
So changing the requirement to 4.0.0 as some 4.0.0 code checks were removed and won't work on previous versions, but should work on 4+.

package.json
```diff
@@ -43,6 +43,10 @@
         "cordova": ">=3.1.0"
       },
       "4.0.0": {
+        "cordova": ">=3.1.0",
```
brodybits, line 46: I think minimum Cordova version should be increased as well, both here and in plugin.xml. I don't think we want to keep supporting old Cordova versions:)

NiklasMerz: See https://lists.apache.org/thread.html/rc277f523ceb625820acd4b361e08b3b0269108f19a4d5e85b9d104fd%40%3Cdev.cordova.apache.org%3E

PR #688, opened 2020-04-29
(ios): allow to set "preferredContentMode"

### LINKED ISSUE #687: InAppBrowser is in desktop mode on iPadOS 13
# Bug Report

## Problem
We are using a framework that detects the device and selects a theme for the OS. It looks like since iPadOS came out iPads are detected as desktop browsers on our webpage.

iPad should identify themselves as iOS devices. The useragent should say something like iPad, iOS like before.

iPads get a macOS like useragent for the InAppBrowser window and an iPad user agent for the main Cordova Webview.

###### Cordova Webview
`Mozilla/5.0 (iPad; CPU OS 13_4_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148`

###### IAB
`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148`

## Information

I found a WWDC talk about changes to the webview Apple is introducing to the iPad: https://developer.apple.com/videos/play/wwdc2019/203/?time=379

**I think we should allow setting the `preferredContentMode` for the IAB webview (and maybe even cordova-ios)**

I think there are some options:

* Set it to mobile to have the same behavior as previos iOS versions
* Add an option to the IAB options
* Add a config.xml preference and use it in IAB and possibly cordova-ios

If we add something like this what would be the default? "recommened" - the iOS default chooses the mode that it thinks are optimal. Setting the default to "mobile" would avoid breaking changes.

iOS/iPadOS

## Checklist

NiklasMerz:
### Platforms affected
ios

### Motivation and Context

Closes #687 

### Description

WIP to show what issue is talking about. See #687 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ At the moment the only supported target in OSX is `_system`.
 
 ### iOS Quirks
 
-Since iOS 13 iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to `Macintosh`. You can change this with the `PreferredContentMode` preference in "config.xml".
+Since the introduction of iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to Macintosh making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the PreferredContentMode preference in "config.xml".
 
 ```xml
 <preference name="PreferredContentMode" value="mobile" />
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ At the moment the only supported target in OSX is `_system`.
 
 ### iOS Quirks
 
-Since the introduction of iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to Macintosh making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the PreferredContentMode preference in "config.xml".
+Since the introduction of iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to Macintosh, making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the `PreferredContentMode` preference in `config.xml`.
 
 ```xml
 <preference name="PreferredContentMode" value="mobile" />
--- a/README.md
+++ b/README.md
@@ -168,6 +168,16 @@ At the moment the only supported target in OSX is `_system`.
 
 `_blank` and `_self` targets are not yet implemented and are ignored silently. Pull requests and patches to get these to work are greatly appreciated.
 
+### iOS Quirks
+
+Since iOS 13 iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to `Macintosh`. You can change this with the `PreferredContentMode` preference in "config.xml".
+
+```xml
+<preference name="PreferredContentMode" value="mobile" />
+```
+
+The example above forces the user agent to contain `iPad`. The other option is to use the value `desktop` to turn the user agent to `Macintosh`.
+
 ### Browser Quirks
 
 - Plugin is implemented via iframe,
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -760,7 +760,7 @@ BOOL isExiting = FALSE;
     
     if (@available(iOS 13.0, *)) {
         NSString *contentMode = [self settingForKey:@"PreferredContentMode"];
-        if ([contentMode  isEqual: @"mobile"]) {
+        if ([contentMode isEqual: @"mobile"]) {
             configuration.defaultWebpagePreferences.preferredContentMode = WKContentModeMobile;
         } else if ([contentMode  isEqual: @"desktop"]) {
             configuration.defaultWebpagePreferences.preferredContentMode = WKContentModeDesktop;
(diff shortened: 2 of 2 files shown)


NiklasMerz:
I think preference totally makes sense and using this preference for the main webview, too. Changed this ways. Let me know what you think.

NiklasMerz:
@dpa99c Thanks for the review. Yes we need to document. Since we also would need that preference in the cordova-ios platform my plan was to do the documentation in cordova-docs where the other prefrences are. The PR for cordova-ios is to be done, too.

jcesarmobile:
I agree with dpa99c, the preference is also going to work on ios and should be documented on ionic-docs for ios, but since it also works in the plugin, should be mentioned in the plugin too

NiklasMerz:
It makes sense to explain this a bit. I pushed an update for README.

NiklasMerz:
Tests are failing probably because CI uses Xcode 10 and this is new for iOS13 in Xcode 11.

README.md
```diff
@@ -168,6 +168,16 @@ At the moment the only supported target in OSX is `_system`.
 
 `_blank` and `_self` targets are not yet implemented and are ignored silently. Pull requests and patches to get these to work are greatly appreciated.
 
+### iOS Quirks
+
+Since iOS 13 iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to `Macintosh`. You can change this with the `PreferredContentMode` preference in "config.xml".
```
dpa99c, line 173: Can I suggest a slightly longer explanation?

Since iOS 13 forked into iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to `Macintosh` making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the `PreferredContentMode` preference in "config.xml".

NiklasMerz:
```suggestion
Since the introduction of iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to Macintosh making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the PreferredContentMode preference in "config.xml".
```

Sounds good to me. However I would try to avoid naming iPadOS an iOS fork.

timbru31:
Did the name change from `iPhone-XR` to 
`		{ platform:iOS Simulator, id:4715F3CB-93A5-46AB-AAEA-D2E86B57C88B, OS:12.1, name:iPhone Xʀ }`?

NiklasMerz:
I dont really know where `iPhone-XR` comes from. I will try different platforms..

jcesarmobile:
I think travis changes should go outside of this PR, probably best to sent a separate PR and once the tests are good in that PR, merge it and rebase this one

erisu:
@timbru31 https://github.com/ios-control/ios-sim/blob/7624cc7af7e875a04f2f7dc3d5c78f2979eb9908/src/helpers.js#L88-L92

erisu:
@NiklasMerz the CI change, ` - env: PLATFORM=local/ios-13.0`, might fail because it does not exist in paramedic, 

https://github.com/apache/cordova-paramedic/tree/master/conf/pr/local

NiklasMerz:
> @NiklasMerz the change in CI configurations might fail because it does not exist in paramedic
> 
> ` - env: PLATFORM=local/ios-13.0`
> 
> https://github.com/apache/cordova-paramedic/tree/master/conf/pr/local

It did fail https://travis-ci.org/github/apache/cordova-plugin-inappbrowser/jobs/694985529. I reverted that.

NiklasMerz:
Looks like I cannot use the Xcode 11 image with this local configuration. paramedic needs to be fixed first.

NiklasMerz:
Is this good to merge? Failing CI is a seperate issue being worked on.

src/ios/CDVWKInAppBrowser.m
```diff
@@ -758,6 +758,15 @@ - (void)createViews
         configuration.mediaPlaybackRequiresUserAction = _browserOptions.mediaplaybackrequiresuseraction;
     }
     
+    if (@available(iOS 13.0, *)) {
+        NSString *contentMode = [self settingForKey:@"PreferredContentMode"];
+        if ([contentMode  isEqual: @"mobile"]) {
```
timbru31, line 763: Objective-C noob here, two spaces required?, some for two lines below

README.md
```diff
@@ -168,6 +168,16 @@ At the moment the only supported target in OSX is `_system`.
 
 `_blank` and `_self` targets are not yet implemented and are ignored silently. Pull requests and patches to get these to work are greatly appreciated.
 
+### iOS Quirks
+
+Since the introduction of iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to Macintosh making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the PreferredContentMode preference in "config.xml".
```
timbru31, line 173:
```suggestion
Since the introduction of iPadOS 13, iPads try to adapt their content mode / user agent for the optimal browsing experience. This may result in iPads having their user agent set to Macintosh, making it hard to detect them as mobile devices using user agent string sniffing. You can change this with the `PreferredContentMode` preference in `config.xml`.
```

timbru31 (changes requested):
two minor bits, otherwise LGTM and I think we can merge

src/ios/CDVWKInAppBrowser.m
NiklasMerz:
```suggestion
        if ([contentMode isEqual: @"mobile"]) {
```

Guess not

timbru31 (approved):

PR #723, opened 2020-06-11
fix(ios): apply PR#672 to uiwebview.

isdh:
### Platforms affected
iOS

### Motivation and Context

I need to use uiwebview on iOS13. In spite of designate landscape mode only, StatusBar somehow orientate as portrait.
I found #672 that resolved wkwebview's StatusBar strange behavior.
but #672 didn't apply to uiwebview. And removed uiwebview on 4.0.x branch.

### Description

apply [#672](https://github.com/apache/cordova-plugin-inappbrowser/pull/672) to uiwebview.

### Checklist

### DIFF
--- a/src/ios/CDVUIInAppBrowser.m
+++ b/src/ios/CDVUIInAppBrowser.m
@@ -259,14 +259,16 @@ static CDVUIInAppBrowser* instance = nil;
     // Run later to avoid the "took a long time" log message.
     dispatch_async(dispatch_get_main_queue(), ^{
         if (weakSelf.inAppBrowserViewController != nil) {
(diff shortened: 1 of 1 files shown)


jcesarmobile:
thanks for the PR, but we don't apply fixes to old versions of plugins, only in the case of security problems, if you need that fix you can create a fork and maintain it for yourself.
Also, UIWebView is deprecated and Apple doesn't allow to use it in new apps since April and won't allow it in existing apps in December, so you should not use it.

isdh:
@jcesarmobile 
OK. thanks for comment. I'll maintain this fix in my repository.

PR #737, opened 2020-06-19
fix(ios): exit event not fired on swipe down

### LINKED ISSUE #718: How to catch event for presentationstyle=pagesheet close by swipe down?
Hi,
I am using presentationstyle = pagesheet options which opens up the browser wonderfully in iOS 13 like a modal but I want to catch the behaviour when a user swipe it down to close the inappbrowser. I added Close button and I can capture that event by calling on 'exit' but what about swipe down to close inappbrowser event?

jcesarmobile:
### Platforms affected
ios

### Motivation and Context

fixes #718

### Description

if presenting IAB on not full screen styles, when the user closes IAB using swipe down gesture instead of the close button the exit event is not fired

exit event is fired on viewDidDisappear, but only if isExiting is set to true, this PR uses the UIAdaptivePresentationControllerDelegate to detect when the modal is being dismissed and set isExiting to true so exit fires on viewDidDisappear

### Checklist

### DIFF
--- a/src/ios/CDVWKInAppBrowser.h
+++ b/src/ios/CDVWKInAppBrowser.h
@@ -49,7 +49,7 @@
 
 @end
 
-@interface CDVWKInAppBrowserViewController : UIViewController <CDVScreenOrientationDelegate,WKNavigationDelegate,WKUIDelegate,WKScriptMessageHandler>{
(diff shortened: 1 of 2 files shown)


kartiksolanki:
Hi,
I tried above changes in my project and while the modal dismiss exit event is fired, I am now facing with UI presentation. So my modal has a close button and, when i do the presentation style to pagesheet (which appears like ios13 modal), the top navigation bar of the modal becomes transparent. I am attaching the image below for your reference.

Screenshot 2020-06-26 at 11 32 18

kartiksolanki:
Any idea what is going wrong in here?

Issue #729, opened 2020-06-16
The html height increased from the second time to open the browser in iOS


lazydan:
# Bug Report

## Problem
At the first time i opened the browser, it worked fine. Then close it and open again. The html height increased.

### Infomation
html {height:100%;}
The First Time:    html.height = screen.height - toolbar.height - statusbar.height
From The Second Time:  html.height = screen.height - toolbar.height

Datas:
iphone  8 (1st):    603px = 667px - 44px - 20px
iphone  8 (from 2nd):    623px = 667px - 44px
iphone 11 (fst):    808px = 896px - 44px - 44px
iphone 11 (from 2nd):    852px = 896px - 44px

### Command or Code
cordova.InAppBrowser.open('https://anydomain.com',  '_blank','location=no,hidenavigationbuttons=yes,toolbarposition=bottom,toolbar=yes,toolbarcolor=#ffffff,hidespinner=yes')

### Environment, Platform, Device
iOS 13.5
iphone 8, iphone 11

### Version information
cordova 9.0.0
cordova-ios 6.0.0
cordova-plugin-inappbrowser 4.0.0
Xcode 11.5

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

chetanbansal019:
++
This issue also exists on iPhone X
Probably related to #301 #306 
Plugin version : 4.0.0 

![Screen Shot 2020-06-17 at 3 50 29 PM](https://user-images.githubusercontent.com/28653223/84887024-d1474e80-b0b2-11ea-9e3a-59065e47161f.png)
![Screen Shot 2020-06-17 at 3 47 41 PM](https://user-images.githubusercontent.com/28653223/84887031-d3a9a880-b0b2-11ea-9e95-6ea45bc2ffdb.png)

giorgos120:
I can confirm that the problem started since the version 4.0.0 upgrade.

This issue, creates serious functionality problems to webpages with content that is  fixed on the bottom of the page. Bellow you can see two images, one with the correct functionality, and one with the new, bug, 
I must point out that on first inappbrowser run, everything works as expected. But if you reopen it, the bug starts.
 
 Without Bug
![download1](https://user-images.githubusercontent.com/2393281/85125562-86eddb00-b234-11ea-8e2b-1f56cada30a4.png)

With Bug
[...]

PR #738, opened 2020-06-19
(ios): Fix incorrect view height from the second open time

lazydan:
Fix incorrect webview height from the second open time.
#729


dpa99c:
I've verified this is using my [test project](https://github.com/dpa99c/cordova-plugin-inappbrowser-test) and it works as expected.

chinieer:
same issue after use last version 4.0.1.when i use camera from file input,after the close the camer,the webview high is incorrect

PR #755, opened 2020-07-31
(android) Removed waitForBeforeload flag as it prevents beforeLoad from being called on every GET request

### LINKED ISSUE #686: beforeload event fires only on even links
# Bug Report

## Problem
the 'beforeload' event fires only on even links, on odd links it doesn’t work

It should fire always

It fire only even (half times)

## Information

Repository to reproduce 
https://github.com/koekaverna/cordova-inapp-testing

### Command or Code

Tested with chome DevTools: chrome://inspect/#devices console output
and android emulator

Pixel_2_API_R:5584

### Version information

Cordova: 9.0.0 (cordova-lib@9.0.1)

## Checklist

PDLMobileApps:
### Platforms affected
Android

### Motivation and Context

Resolves #686 

Link to the issue: #686 

### Description

We removed waitForBeforeload flag as it prevents beforeLoad from being called on every GET request and it does not seem to have any meaningful use.

### Testing

Testing was done manually as I am not sure how to automate it.

The following tests were executed for Android 7, 8, and 10.

Created a new Cordova test app using cordova-plugin-inappbrowser.
Executed:
cordova.InAppBrowser.open("https://www.hannaford.com", "_blank", "beforeload=get");
and verified beforeLoad gets called with every get request.

Executed:
cordova.InAppBrowser.open("mailto:a@hannaford.com", "_blank", "beforeload=get");
and verified beforeLoad gets called with that get request.

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -269,12 +269,6 @@ public class InAppBrowser extends CordovaPlugin {
                 @SuppressLint("NewApi")
                 @Override
                 public void run() {
-                    if (android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.O) {
-                        currentClient.waitForBeforeload = false;
-                        inAppWebView.setWebViewClient(currentClient);
-                    } else {
-                        ((InAppBrowserClient)inAppWebView.getWebViewClient()).waitForBeforeload = false;
-                    }
                     inAppWebView.loadUrl(url);
                 }
             });
@@ -1163,7 +1157,6 @@ public class InAppBrowser extends CordovaPlugin {
         EditText edittext;
         CordovaWebView webView;
         String beforeload;
-        boolean waitForBeforeload;
 
         /**
          * Constructor.
@@ -1175,7 +1168,6 @@ public class InAppBrowser extends CordovaPlugin {
             this.webView = webView;
             this.edittext = mEditText;
             this.beforeload = beforeload;
-            this.waitForBeforeload = beforeload != null;
         }
 
         /**
@@ -1236,7 +1228,7 @@ public class InAppBrowser extends CordovaPlugin {
             }
 
             // On first URL change, initiate JS callback. Only after the beforeload event, continue.
-            if (useBeforeload && this.waitForBeforeload) {
+            if (useBeforeload) {
                 if(sendBeforeLoad(url, method)) {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


dpa99c:
@wvengen originally implemented the `beforeload` feature and [his PR comment](https://github.com/apache/cordova-plugin-inappbrowser/pull/276/files#r233077431) indicates that `waitForBeforeload` is necessary.

have you got an opinion on this @wvengen?

PDLMobileApps:
> @wvengen originally implemented the `beforeload` feature and [his PR comment](https://github.com/apache/cordova-plugin-inappbrowser/pull/276/files#r233077431) indicates that `waitForBeforeload` is necessary.
> 
> have you got an opinion on this @wvengen?

We saw it was added on purpose and I believe it may had a specific goal. However, now it has the effect of preventing beforeLoad from being called on every GET request, which is quite a big deal.
We removed it and we have been using the code in this PR successfully in production for a week.

stionic:
waitForBeforeload is necessary. I think it prevent duplicate event.

for fix problem you maybe want correct it instead of remove.

I have found same problem before and fixed by correct it when onPageFinished. Problem is when load error it not reset waitForBeforeload.

Could you please check [commit](https://github.com/stionic/cordova-plugin-inappbrowser/commit/d2ddbfe37ceceb9ace56007c282902e765e41076)?

PDLMobileApps:
> waitForBeforeload is necessary. I think it prevent duplicate event.
> 
> for fix problem you maybe want correct it instead of remove.
> 
> I have found same problem before and fixed by correct it when onPageFinished. Real problem is when load error it not reset waitForBeforeload.
> 
> Could you please check [commit](https://github.com/stionic/cordova-plugin-inappbrowser/commit/d2ddbfe37ceceb9ace56007c282902e765e41076)?

We have been using the change in this PR in our app in the Google Play Store for almost two month (500k installs). We have no reports of issues with beforeLoad which is called a lot when the customer uses the app. So, if waitForBeforeload is really necessary, I would expect a completely different situation.

By the way, no one has been able to provide a good explanation on why it's needed. On the other hand, if you have in the code, beforeLoad only happens every other request, which is definitely wrong. That has been reported more than once by many people.

dtarnawsky:
To follow up on this PR, I've found in testing that this PR is necessary for beforeload to function more than once. @stionic:  The code for waitForBeforeload doesn't prevent duplicates of the beforeload event occurring and can be safely removed  .

ep-mark:
This PR fixes the issues we found too!

wunschradio:
Thank you. This finally fixed this issue for me.

PR #778, opened 2020-09-09
fix(ios):  Allow loading of external local files from the same directory as loaded local html

### LINKED ISSUE #657: Unable to load a local URL or html file in Corodva iOS with inAppBrowser
# Bug Report

I have recently upgraded to Cordova InAppBrowser plugin to 3.2.0 which uses wkwebview to render HTML content. I have always been able to load local HTML file by using the following the code:

```
 cordova.InAppBrowser.open(
          'cdvfile://localhost/bundle/www/assets/index.html',
          '_blank',
        );
```

However, wkwebview doesn't support 'cdvfile' so I tried using the relative file as below

```
 cordova.InAppBrowser.open(
          'assets/index.html',
          '_blank',
        );
```

This approach is not working either. The inAppBrowser opens and shows HTML Error 403: forbidden. After I close the browser, all the local files like IMG becomes unavailable and browser console [Error] Failed to load resource: the server responded with a status of 403 (Forbidden). I don't understand why closing browser cause other local files become accessible.

I m using the cordova-plugin-wkwebview-engine plugin which forces to use WKwebview. here are the configs

```
<preference name="WKWebViewOnly" value="false" />
        <feature name="CDVWKWebViewEngine">
            <param name="ios-package" value="CDVWKWebViewEngine" />
        </feature>
 <preference name="CordovaWebViewEngine" value="CDVWKWebViewEngine" />
```

### Version information

Cordova ios: 5.1.1
ios: 13.4
Cordova: 9.0

## Checklist

- [ ] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

guycalledseven:
# Platforms affected

iOS

### Motivation and Context

Many applications rely on ability to display local resources using inappbrowser. 
WKWebView implementation introduced additional security which needs to be handled.

Fixes #657 
Improves on #883 

### Description
Pull request #693 made loading of local html files is possible in v4 WKWebView branch. However - how it was implemented only enabled opening of a single html file in IAB without its external resources (eg. css/images/javascript) throwing `didStartProvisionalNavigation` (not allowed).

As per [Apple documentation](https://developer.apple.com/documentation/webkit/wkwebview/1414973-loadfileurl) `loadFileURL` accepts 2 parameters:
- `URL` - The file URL to navigate to.
- `readAccessURL` -The URL to allow read access to.

If readAccessURL references a single file, only that file may be loaded by WebKit. 
**If readAccessURL references a directory, files inside that file may be loaded by WebKit.**

Now, I extract directory from passed url and pass it to the  `loadFileURL` method making **external local content from the same directory as loaded html** - loadable.

### Testing

npm test pass. multiple builds on multiple real devices with and without notch.

### Checklist

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -1088,7 +1088,8 @@ BOOL isExiting = FALSE;
 - (void)navigateTo:(NSURL*)url
 {
     if ([url.scheme isEqualToString:@"file"]) {
-        [self.webView loadFileURL:url allowingReadAccessToURL:url];
+        NSURL* directoryURL = [url URLByDeletingLastPathComponent];
+        [self.webView loadFileURL:url allowingReadAccessToURL:directoryURL];
     } else {
         NSURLRequest* request = [NSURLRequest requestWithURL:url];
         [self.webView loadRequest:request];


regnete:
Any hint about when will this be merged? Its a show stopper for our migration to cordova-ios 6 / wkwebview. Thx.

timbru31 (commented):
LGTM to me; however IMO we should test this. Do you have an example app that could speed up testing this PR?

guycalledseven:
Yes, I have 4 working apps in production via AppStore. They passed Apple. 😅

jcesarmobile:
When we ask for "example apps" we mean a github repository of a cordova app (or zip file) where the issue can be reproduced and then can be used to verify the fix.

src/ios/CDVWKInAppBrowser.m
```diff
@@ -1097,7 +1097,8 @@ - (void)close
 - (void)navigateTo:(NSURL*)url
 {
     if ([url.scheme isEqualToString:@"file"]) {
-        [self.webView loadFileURL:url allowingReadAccessToURL:url];
+        NSURL* directoryURL = [url URLByDeletingLastPathComponent];
+        [self.webView loadFileURL:url allowingReadAccessToURL:directoryURL];
```
breautek, line 1101: Would like allow reading the local `cordova.js` file? If so that could be **big** trouble...

guycalledseven: This does not change current logic anyhow. This only enables later ios versions to load static files from `url`.

Use case is - we put static content in our apps and then app reads this static content via built in webserver / directly (css files/images etc).

guycalledseven:
> When we ask for "example apps" we mean a github repository of a cordova app (or zip file) where the issue can be reproduced and then can be used to verify the fix.

This is so easily reproduced. I really cannot believe it how long it took for someone to notice this PR.
There is so many IOS issues related to this in Github Issues for this project.

lehizojie:
I am having this problem too.
With this fix are we supposed to be able to use cordova.InAppBrowser.open('help/signin.html', '_blank'), where help is a folder sibling to index.html?
Trying to get this to work with no luck. Any help would be appreciated.

src/ios/CDVWKInAppBrowser.m
sven513: This fix is really a lifesaver for accessing local resources.
Fixes "The operation couldn’t be completed. (kCFErrorDomainCFNetwork error 1.)" error for me, and saved my clients apps!
Any chance this gets committed any time soon?

PR #788, opened 2020-09-22
Keyboard Dismissal Leaves Viewport Shifted in iOS < 13.4. (3.2.X)

AUK4GIT:
apache/cordova-ios#417

### Checklist


timbru31 (commented):
I doubt we will merge this workaround for an Apple bug.

dpogue:
We definitely don't want to introduce any programmatic scrolling, because it can cause the page to jump around during interaction and be very disorienting for users.

You might be able to use https://www.npmjs.com/package/cordova-plugin-wkkeyboardfix (which should solve the problem for all WKWebViews in an app, regardless of whether they are the Cordova webview or an In-App-Browser webview)

Issue #554, opened 2019-10-15
InAppBrowser.java:268:error: cannot find symbol


supersongpiao:
# Bug Report

## Problem
The plugin version I am using is 3.1.1-dev.
![WeChat9f2b4aa66dd21080b77ff4090f372313](https://user-images.githubusercontent.com/31645741/66811611-facc3b00-ef63-11e9-8dd7-57df5f78a3f7.png)

### Version information
Ionic:

   Ionic CLI         : 5.2.1 (/usr/local/lib/node_modules/ionic)
   Ionic Framework   : ionic1 1.3.2
   @ionic/v1-toolkit : not installed

Cordova:

   Cordova CLI       : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms : android 6.2.3
   Cordova Plugins   : cordova-plugin-ionic-webview 1.2.1, (and 26 other plugins)

Utility:

   cordova-res : 0.6.0 (update available: 0.8.0)
   native-run  : 0.2.7 (update available: 0.2.8)

System:

   Android SDK Tools : 26.1.1 (/Users/rdc-002/Library/Android/sdk)
   ios-deploy        : 2.0.0
   ios-sim           : 6.1.2
   NodeJS            : v10.15.3 (/usr/local/bin/node)
   npm               : 6.10.0
   OS                : macOS Mojave
   Xcode             : Xcode 11.1 Build version 11A1027

## Checklist

- [ ] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

jcesarmobile:
The problem is you are using a very old version of cordova-Android, but the plugin should require that version, so should be fixed in the plugin engines so it doesn’t install in older versions

supersongpiao:
> The problem is you are using a very old version of cordova-Android, but the plugin should require that version, so should be fixed in the plugin engines so it doesn’t install in older versions

Yes, I found out later. This version of plug-in requires Android API 26 or above. Can I know when the official version of 3.1.1 will be released?

jcesarmobile:
I'm reopening because it's a bug, not that it doesn't work in older version, but that the engines in plugin.xml and package.json are not properly configured to let cordova know it's not compatible with older version

ejonnadula:
I am also facing same issue, Is there any fix/solution available 

[...]

PR #803, opened 2020-10-20
fix(android): allow compilation in old cordova-android versions

jcesarmobile:
since the plugin don't have engines for cordova-android, people is using it in old cordova-android versions and it fails to compile because it uses some code that is not supported in older SDK compile versions.

This PR should make it compile in older versions (tested only 6.2.3 which targets android-25, could still fail in even older versions)

closes https://github.com/apache/cordova-plugin-inappbrowser/issues/554

after this is merged we should do a patch release and then we can revert and set the engines to cordova-android 9 and remove a lot of code that wouldn't be needed, and then do a major release as updating the engines is a breaking change

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -269,12 +269,8 @@ public class InAppBrowser extends CordovaPlugin {
                 @SuppressLint("NewApi")
                 @Override
                 public void run() {
-                    if (android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.O) {
-                        currentClient.waitForBeforeload = false;
-                        inAppWebView.setWebViewClient(currentClient);
-                    } else {
-                        ((InAppBrowserClient)inAppWebView.getWebViewClient()).waitForBeforeload = false;
-                    }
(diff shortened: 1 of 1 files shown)


breautek (approved):

breautek:
I'm assuming this will cause breakage if running on later android targets. We should probably attempt to get this patch and the new major released around the same time.

jcesarmobile:
no, it should still work on any capacitor-android version, even on newer ones, it just replaces constants which might not exist with the number they represent
and this other change https://github.com/apache/cordova-plugin-inappbrowser/pull/803/files#diff-b73be6c976a963c35af9c94c6c85447ca3fa5a70f417912f71cce92b16896874L272-L277 just removes `getWebViewClient` usage, which is not available if you compile on cordova-android 6.2.3 (SDK 25) or older, but the code does the same thing.

timbru31 (commented):
Not really a big fan of introducing magic numbers, the Android constants exists for a reason and make reading the code a lot easier.

jcesarmobile:
Yeah, this is a temporary fix, will revert it once released

erisu (approved):

PR #821, opened 2020-11-22
breaking(android): replace magic numbers with android.os.Build constants

jcesarmobile:
Breaking because it won't compile in old cordova-android versions, engines should be updated to only allow cordova-android 9 and a major bump should be done on next release

Reverts apache/cordova-plugin-inappbrowser#803

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -269,8 +269,12 @@ public class InAppBrowser extends CordovaPlugin {
                 @SuppressLint("NewApi")
                 @Override
                 public void run() {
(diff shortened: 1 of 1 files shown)


timbru31 (commented):
If we go with a breaking change, we should be safe to delete some safeguard? - as we agreed to support only Android 5.1 and higher (some code checks for, e.g., 17/Jellybean or >= 21/Lollipop.

jcesarmobile:
Yeah, this is just a revert of the previous commit where the magic numbers were introduced.
I plan to do some code cleanup in future PRs

PR #823, opened 2020-11-24
chore: bump engines requirements

jcesarmobile:
require newer cordova-ios, cordova-android and cordova versions

PR #824, opened 2020-11-24
breaking: cleanup old android versions code

jcesarmobile:
if we target cordova-android 9 as proposed on https://github.com/apache/cordova-plugin-inappbrowser/pull/823, we can remove all this old code that wouldn't be used anymore


timbru31 (approved):
LGTM codewise

PR #825, opened 2020-11-25
(ios): rename CDVWKProcessPoolFactory

### LINKED ISSUE #751: WKWebview cookie sometimes can't be get after upgrade to Cordova 6.1.0
# Bug Report

## Problem
Try to get the cookie of inappbrowser. But sometimes can't get correct cookie.
This issue happen after upgrade cordova to Cordova ios 6.1.0. and inappbrowser upgrade to 4.0.0

Use  blow code to get cookies in decidePolicyForNavigationAction of CDVWKInAppBrowser

WKHTTPCookieStore *wkWebViewCookieStore =[[[theWebView configuration] websiteDataStore] httpCookieStore];
 [wkWebViewCookieStore getAllCookies:^(NSArray* wkcookies) {
        for (NSHTTPCookie* cookie in wkcookies)
               {
                   NSLog(@"wk cookie name is %@",cookie.name);
               }
    }];

Some times the wkcookies can't get the correct cookie.
And this issue not always happen,  but sometimes.   in my test ,50% percentage happen.

## Information

This issue not happen after downgrade the cordova ios to 5.1 and still use inappbrowser 4.0.0(after resolve some compile error).

### Command or Code

Please see above 

Cordova 9.0.0, Cordova ios 6.1.0, xcode 11.5. 
Iphone simulator(13.5) and Iphone X (13.5)

## Checklist

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

CDVWKProcessPoolFactory was integrated from the WKWebView plugin with
the new name: CDVWebViewProcessPoolFactory

### Description

This fixes the import of the processpool factory for cordova-ios versions 6 or greater. The process pool is now part of cordova-ios and needs to be available public see: https://github.com/apache/cordova-ios/pull/1031

Closes #751 

### Checklist

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -19,8 +19,10 @@
 
 #import "CDVWKInAppBrowser.h"
 
-#if __has_include(<Cordova/CDVWebViewProcessPoolFactory.h>)
-#import <Cordova/CDVWebViewProcessPoolFactory.h>
+#if __has_include(<Cordova/CDVWebViewProcessPoolFactory.h>) // Cordova-iOS >=6
+  #import <Cordova/CDVWebViewProcessPoolFactory.h>
+#elif __has_include("CDVWKProcessPoolFactory.h") // Cordova-iOS <6 with WKWebView plugin
+  #import "CDVWKProcessPoolFactory.h"
 #endif
 
 #import <Cordova/CDVPluginResult.h>
@@ -741,6 +743,8 @@ BOOL isExiting = FALSE;
     configuration.userContentController = userContentController;
 #if __has_include(<Cordova/CDVWebViewProcessPoolFactory.h>)
     configuration.processPool = [[CDVWebViewProcessPoolFactory sharedFactory] sharedProcessPool];
+#elif __has_include("CDVWKProcessPoolFactory.h")
+    configuration.processPool = [[CDVWKProcessPoolFactory sharedFactory] sharedProcessPool];
 #endif
     [configuration.userContentController addScriptMessageHandler:self name:IAB_BRIDGE_NAME];
     
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -19,8 +19,8 @@
 
 #import "CDVWKInAppBrowser.h"
 
-#if __has_include("CDVWKProcessPoolFactory.h")
-#import "CDVWKProcessPoolFactory.h"
+#if __has_include(<Cordova/CDVWebViewProcessPoolFactory.h>)
+#import <Cordova/CDVWebViewProcessPoolFactory.h>
 #endif
 
 #import <Cordova/CDVPluginResult.h>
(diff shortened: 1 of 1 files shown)


src/ios/CDVWKInAppBrowser.m
```diff
@@ -19,8 +19,8 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 #import "CDVWKInAppBrowser.h"
 
-#if __has_include("CDVWKProcessPoolFactory.h")
-#import "CDVWKProcessPoolFactory.h"
+#if __has_include(<Cordova/CDVWebViewProcessPoolFactory.h>)
```
dpogue, line 22: In theory, for maximum compatibility, you could try both options:

```
#if __has_include(<Cordova/CDVWebViewProcessPoolFactory.h>) // Cordova-iOS >=6
  #import <Cordova/CDVWebViewProcessPoolFactory.h>
#elif __has_include("CDVWKProcessPoolFactory.h") // Cordova-iOS <6 with WKWebView plugin
  #import "CDVWKProcessPoolFactory.h"
#endif
```

NiklasMerz: That makes sense since the requirements are just set to cordova-ios>5.1.0 in #823

timbru31 (approved):
LGTM, but obviously only if we merge in https://github.com/apache/cordova-ios/pull/1031 :)

timbru31:
With Darryl's suggestion the milestone could be moved to a minor release I guess?

NiklasMerz:
Yes the change is not breaking anymore. Looks like the next release will be major anyways and we don't have a minor milestone.

erisu (approved):

dpogue (approved):

jcesarmobile:
since 4.1.0 introduced [breaking changes](https://github.com/apache/cordova-plugin-inappbrowser/issues/831) maybe we should do a minor release first that fixed that problem and then we can continue with the breaking changes.
But that would mean we need to rever the [merged breaking changes](https://github.com/apache/cordova-plugin-inappbrowser/pull/821) to be able to do a minor release

or we can continue with the major release as not supporting Xcode 10 is not a big issue since its builds are not accepted in the App Store

NiklasMerz:
Yeah I was just thinking about when and how to release next. Maybe we should do a discuss thread on the list.

Personally I think I would go for a major release now. I would agree that supporting Xcode 10 should not be that important since Xcode 11 is required for the app store and the current release is even 12. We should reduce our workload and move forward with the limited time we have. 

@jcesarmobile Do you know issues or users about Xcode 10?

jcesarmobile:
> @jcesarmobile Do you know issues or users about Xcode 10?


Don't understand the question, the issue on Xcode 10 is linked on my previous comment https://github.com/apache/cordova-plugin-inappbrowser/issues/831

The easy fix for users is to use Xcode 11 or newer to compile, if we wanted to support it a `#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000` macro should be used, but probably not worth it

NiklasMerz:
I was just asking if you know about people having issues with this version. I personally always prioritize work if users are complaining and it's not just something that came up internally.

jcesarmobile:
I created the issue and nobody has added the 👍  reaction to it, so I guess nobody was affected by it.

NiklasMerz:
As this fix needs this PR to be released with cordova-ios https://github.com/apache/cordova-ios/pull/1031  I think we could wait for that or are there reasons to release inappbrowser now as 5.0.0? The milestone looks good.

nmanikiran:
Hi @NiklasMerz 
i am facing issues with cookie sync between `wkwebview` & `InappBrowser` even after updating to `v5.0.0`
for more i logged an [issue](https://github.com/apache/cordova-plugin-inappbrowser/issues/877)

PR #765, opened 2020-08-18
(ios) Add InAppBrowserStatusBarStyle preference

mrbberra:
### Platforms affected
iOS

### Motivation and Context
This allows apps that set a dark status bar background color using [cordova-plugin-statusbar](https://github.com/apache/cordova-plugin-statusbar) to set the status bar style when using the in app browser. Became necessary after [GH-656](https://github.com/apache/cordova-plugin-inappbrowser/pull/656).
Fixes: #728 

### Description
Adds the ability to set `<preference name="InAppBrowserStatusBarStyle" value="lightcontent" />` in the config.xml file, so that the `preferredStatusBarStyle` becomes `UIStatusBarStyleLightContent` instead of `UIStatusBarStyleDefault`.

### Testing
I created a [test app](https://github.com/mrbberra/inappbrowser-statusbar-style) and tried setting the InAppBrowserStatusBarStyle to "default", "lightcontent", and removing it alltogether. I didn't add any new automatic tests- not sure how.

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x] I've updated the documentation if necessary

PR #828, opened 2020-11-30
feat(ios): add InAppBrowserStatusBarStyle 'darkcontent' configuration option

NiklasMerz:
### Platforms affected
iOS

### Motivation and Context

If the IAB loads a page with a white background on a device with dark mode enabled the statusbar will be white with white content. This PR ads the config.xml to set statusbar content to dark to make it readable again for pages like this.

### Description

https://github.com/apache/cordova-plugin-inappbrowser/pull/765 added the option to set the statusbar to light content. iOS 13 introduced a new dark content option and this PR adds that.

### Testing

Manual test
No automatic tests

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ simply hook `window.open` during initialization.  For example:
 ### Preferences
 
 #### <b>config.xml</b>
-- <b>InAppBrowserStatusBarStyle [iOS only]</b>: (string, options 'lightcontent', 'darkcontent' or 'default'. Defaults to 'default') set text color style for iOS.
+- <b>InAppBrowserStatusBarStyle [iOS only]</b>: (string, options 'lightcontent', 'darkcontent' or 'default'. Defaults to 'default') set text color style for iOS. 'lightcontent' is intended for use on dark backgrounds. 'darkcontent' is only available since iOS 13 and intended for use on light backgrounds.
 ```
 <preference name="InAppBrowserStatusBarStyle" value="lightcontent" />
 ```
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ simply hook `window.open` during initialization.  For example:
 ### Preferences
 
 #### <b>config.xml</b>
-- <b>InAppBrowserStatusBarStyle [iOS only]</b>: (string, options 'lightcontent' or 'default'. Defaults to 'default') set text color style for iOS.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -71,7 +71,7 @@ simply hook `window.open` during initialization.  For example:
 ### Preferences
 
 #### <b>config.xml</b>
-- <b>InAppBrowserStatusBarStyle [iOS only]</b>: (string, options 'lightcontent' or 'default'. Defaults to 'default') set text color style for iOS.
+- <b>InAppBrowserStatusBarStyle [iOS only]</b>: (string, options 'lightcontent', 'darkcontent' or 'default'. Defaults to 'default') set text color style for iOS.
```
jcesarmobile, line 74: Should probably mention that darkcontent is iOS 13+ only, and a bit more info about the styles could be helpful too

NiklasMerz: I checked Apples API docs and added a description.

src/ios/CDVWKInAppBrowser.m
```diff
@@ -1067,6 +1067,12 @@ - (UIStatusBarStyle)preferredStatusBarStyle
     NSString* statusBarStylePreference = [self settingForKey:@"InAppBrowserStatusBarStyle"];
     if (statusBarStylePreference && [statusBarStylePreference isEqualToString:@"lightcontent"]) {
         return UIStatusBarStyleLightContent;
+    } else if (statusBarStylePreference && [statusBarStylePreference isEqualToString:@"darkcontent"]) {
+        if (@available(iOS 13.0, *)) {
+            return  UIStatusBarStyleDarkContent;
```
timbru31, line 1072:
```suggestion
            return UIStatusBarStyleDarkContent;
```

a bit nitpicky, but there is a space too much.

NiklasMerz: Good catch

jcesarmobile:
We should be careful with the `@available` usage, the code inside it is not understand if using an older SDK to build than the version on the `@available`, which means, this code doesn't compile on Xcode 10 or older.

For this PR is not a big deal since we are planning a major release and can bump the engines to cordova-ios 6 instead of 5.1.0 as on my pending PR, but then noticed version 4.1.0 is already using `@available` with iOS 13 and fails to build.

I've reported [the issue](https://github.com/apache/cordova-plugin-inappbrowser/issues/831)

jcesarmobile (approved):

PR #882, opened 2021-06-16
(android) Fix for webView window not being destroyed correctly causing memory leak (#290)

### LINKED ISSUE #290: InAppBrowser will not close when invoking ref.close()
On both Android and iOS, ref.close() does not close the window.  I can see the webview is still running.  Does not seem to be specific to device or OS.  However, I have been testing on iPhone X, with iOS 11.4.1 as well as Samsung S5 running Android 6.0.1.

PDLMobileApps:
### Platforms affected
android

### Motivation and Context

This PR fixes #290 

### Description

A fix for the issue reported in #290 is suggested in the same. I created this pull request after implementing the fix in my own forked repo and verifying it works as intended in Android.

### Testing

- Launched the app compiled without the fix on Android and verified the issue is present using the Chrome Inspector (window with about:blank is present after IAB window is closed)
- Compiled the app with the fix and launched it
- Opened an IAB window
- Closed the IAB window and verified the issue is no longer present using the Chrome Inspector (window with about:blank is no longer present after IAB window is closed)
- Repeated the test a few time

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -534,6 +534,16 @@ public class InAppBrowser extends CordovaPlugin {
                             dialog.dismiss();
                             dialog = null;
                         }
+                        
+                        // Fix for webView window not being destroyed correctly causing memory leak
+                        // (https://github.com/apache/cordova-plugin-inappbrowser/issues/290)
+                        if (url.equals(new String("about:blank"))) {
+                            inAppWebView.onPause();
+                            inAppWebView.removeAllViews();
+                            inAppWebView.destroyDrawingCache();
+                            inAppWebView.destroy();
+                            inAppWebView = null;
+                        }
                     }
                 });
                 // NB: From SDK 19: "If you call methods on WebView from any thread


mosabab (approved):
The code LGT
Thanks for contributing

src/android/InAppBrowser.java
```diff
@@ -534,6 +534,16 @@ public void onPageFinished(WebView view, String url) {
                             dialog.dismiss();
                             dialog = null;
                         }
+                        
+                        // Fix for webView window not being destroyed correctly causing memory leak
+                        // (https://github.com/apache/cordova-plugin-inappbrowser/issues/290)
+                        if (url.equals(new String("about:blank"))) {
```
ndrake, line 540: The call to `new String()` is unnecessary, consider doing: `if (url.equals("about:blank")) {`

marc7000:
When will the next version with these fixes be released?

ilyakamens:
Hi @NiklasMerz, when will `5.next` land?

PR #883, opened 2021-06-17
fix(android): Allow loading local files

NoelDeMartin:
### Platforms affected

Android

### Motivation and Context

This PR allows opening local files in the inappbrowser for Android applications targeting SDK 30 or later. The use-case for this is the same as #693, but applicable to Android instead of iOS.

### Description

Starting with SDK 30, the default value for the `allowFileAccess` setting changed in Android (as [specified in the documentation](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean))). It was `true` before and it's `false` now, meaning that this can potentially break existing applications.

This problem is something that has already been tackled in cordova core (Initially in https://github.com/apache/cordova-android/pull/1111, later updated in https://github.com/apache/cordova-android/pull/1222). So I just followed the same strategy and this setting is only updated if the `AndroidInsecureFileModeEnabled` preference is set to `true`.

### Testing

I've tested the fix in my application. Given that the code is quite simple and it's already in cordova core, I don't think it needs any further testing.

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -945,6 +945,13 @@ public class InAppBrowser extends CordovaPlugin {
                 settings.setBuiltInZoomControls(showZoomControls);
                 settings.setPluginState(android.webkit.WebSettings.PluginState.ON);
 
+                if (preferences.getBoolean("AndroidInsecureFileModeEnabled", false)) {
+                    LOG.d(LOG_TAG, "Enabled insecure file access");
+
+                    settings.setAllowFileAccess(true);
+                    settings.setAllowUniversalAccessFromFileURLs(true);
+                }
+
                 // Add postMessage interface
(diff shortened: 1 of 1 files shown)


quenis:
Hello, please accept the merge 'cause we need that correction. Thanks bro!

NoelDeMartin:
@quenis Until this gets merged, you can use our fork which is the 5.0 release + this one fix. https://github.com/moodlemobile/cordova-plugin-inappbrowser

quenis:
> @quenis Until this gets merged, you can use our fork which is the 5.0 release + this one fix. https://github.com/moodlemobile/cordova-plugin-inappbrowser

@NoelDeMartin Thanks man for yr reply!

torsorensen (approved):

Issue #885, opened 2021-06-23
[Android] Inappbrowser cannot handle the intent links with fallback URL


mosabab:
# Bug Report

## Problem
[Android] Inappbrowser cannot handle the intent links with fallback URL

### What is expected to happen?
In inappbrowser, If the user click on an intent link with fallback URL and when an intent could not be resolved, or an external application could not be launched, then the user will be redirected to the fallback URL if it was given.

### What does actually happen?
If the user click on an intent link with fallback URL and when an intent could not be resolved, or an external application could not be launched, the inappbrowser cannot handle the fallback url and the webview generate the following error messagge:
Could not load because of net::ERR_UNKNOWN_URL_SCHEME

## Information

A little known feature in Android lets you launch apps directly from a web page via an Android Intent. One scenario is launching an app when the user lands on a page, which you can achieve by embedding an iframe in the page with a custom URI-scheme set as the src, as follows: `<iframe src="paulsawesomeapp://page1"> </iframe>`. This works in the Chrome for Android browser, version 18 and earlier. It also works in the Android browser, of course.

The functionality has changed slightly in Chrome for Android, versions 25 and later. It is no longer possible to launch an Android app by setting an iframe's src attribute. For example, navigating an iframe to a URI with a custom scheme such as `paulsawesomeapp://` will not work even if the user has the appropriate app installed. Instead, you should implement a user gesture to launch the app via a custom scheme, or use the "intent:" syntax described in this article.

Syntax 
The best practice is to construct an intent anchor and embed that into the page so the user can launch the app. This gives you a lot more flexibility in controlling how apps are launched, including the ability to pass extra information into the app via Intent Extras.

The basic syntax for an intent-based URI is as follows:

`intent:  
   HOST/URI-path // Optional host  
   #Intent;  
      package=\[string\];  
      action=\[string\];  
      category=\[string\];  
      component=\[string\];  
      scheme=\[string\];  
   end;`

Also, you may choose to specify fallback URL by adding the following string extra:

`S.browser_fallback_url=[encoded_full_url]`
When an intent could not be resolved, or an external application could not be launched, then the user will be redirected to the fallback URL if it was given.

Some example cases where Chrome does not launch an external application are as follows:

* The intent could not be resolved, i.e., no app can handle the intent.
* JavaScript timer tried to open an application without user gesture.

[...]

Issue #880, opened 2021-06-03
"intent:" URI is parsed incorrectly


t4kemyh4nd:
https://github.com/apache/cordova-plugin-inappbrowser/blob/3b6757fc4407aa5cd9d5cbda3ff2f8174ecdae1a/src/android/InAppBrowser.java#L1203

In the shouldOverrideUrlLoading, the code seems to be also capable of parsing intent URIs. However, in case of intent URI, for example **intent://scan/#Intent;scheme=zxing;package=com.google.zxing.client.android;end**, `intent.setData` will not work at all when it is passed to `startActivity`. To be able to parse intent URIs, `intent.parseUri` should be used instead.

t4kemyh4nd:
Exact same issue highlighted here: https://stackoverflow.com/questions/33151246/how-to-handle-intent-on-a-webview-url

mosabab:
Hi
I don't know if your post related to this issue https://github.com/apache/cordova-plugin-inappbrowser/issues/867 ?

mosabab:
> Exact same issue highlighted here: https://stackoverflow.com/questions/33151246/how-to-handle-intent-on-a-webview-url

if you found the solution, please share your fix via PR.

Thanks

mosabab:
@breautek 
Hi, Sorry for mentioned ,
could you please share your idea if can fix this issue.

Thanks

mosabab:
> Exact same issue highlighted here: https://stackoverflow.com/questions/33151246/how-to-handle-intent-on-a-webview-url

@t4kemyh4nd 
Do you have the correct code for to fix inappbrowser plugin?
If you know please share it or do it in PR.
Thanks

kev494:
Can we expect a fix for this issue soon? It still persists and opening the tabs with `_system` instead of `_blank` is not really an option for all use cases.

PR #886, opened 2021-06-27
[Android] Inappbrowser: ability to handles intent scheme links

mosabab:
### Platforms affected
Android

### Motivation and Context

closes https://github.com/apache/cordova-plugin-inappbrowser/issues/885
closes https://github.com/apache/cordova-plugin-inappbrowser/issues/880

### Description

This PR let Inappbrowser plugin to handle the intents,
If the link is an INTENT then handle as:
1) Open the associated app (if it installed).
or
2) If the app isn't installed on the device, then try to open the fallback URL (if it given in the URL).
or
3) Go to the Store app link.

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -20,6 +20,7 @@ package org.apache.cordova.inappbrowser;
 
 import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
@@ -29,6 +30,8 @@ import android.provider.Browser;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
+import android.graphics.PorterDuff;
+import android.graphics.PorterDuffColorFilter;
 import android.graphics.Color;
 import android.net.http.SslError;
 import android.net.Uri;
@@ -45,6 +48,7 @@ import android.view.WindowManager.LayoutParams;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputMethodManager;
 import android.webkit.CookieManager;
+import android.webkit.CookieSyncManager;
 import android.webkit.HttpAuthHandler;
 import android.webkit.JavascriptInterface;
 import android.webkit.SslErrorHandler;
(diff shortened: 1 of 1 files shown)


src/android/InAppBrowser.java
```diff
@@ -996,9 +996,6 @@ public void postMessage(String data) {
                 inAppWebView.setId(Integer.valueOf(6));
                 inAppWebView.getSettings().setLoadWithOverviewMode(true);
                 inAppWebView.getSettings().setUseWideViewPort(useWideViewPort);
-                // Multiple Windows set to true to mitigate Chromium security bug.
-                //  See: https://bugs.chromium.org/p/chromium/issues/detail?id=1083819
-                inAppWebView.getSettings().setSupportMultipleWindows(true);
```
jcesarmobile, line 1001: this is here to mitigate a Chromium security bug as the comment explained, so you are reintroducing the issue again

jcesarmobile (changes requested):

src/android/InAppBrowser.java
mosabab: This bug has been closed, please refer to the link

mosabab: Also this line prevent intent links from work

mosabab: I try to check intent links, but it not working because of this line you mentioned

jcesarmobile: It's closed because it was fixed in newer chromium versions, but we can't guarantee the user has the patched version, so we need this code to prevent the attack. If it also prevent the intents, it can probably be fixed in a different way.

breautek: The flaw applies to all versions of chrome webview prior to 83.0.4103.106.

Older android devices may not be able to run the latest webview which would contain the security patch (I know android 5.1 for example only have version 76 available on the google play store).

These android versions are still versions Cordova officially supports, which is why this mitigation line is necessary.

https://nvd.nist.gov/vuln/detail/CVE-2020-6506#range-6218332

If this line blocks the feature to handle intent links, then we cannot support the feature at this time. If you need this feature, you fork this plugin, but I would recommend increasing your min sdk to a level where you can ensure that all your users can be running the patched webview.

mosabab: You are welcome to improve this PR in the correct way 👍

mosabab: You said we can fix in other way, so we need to keep that line also we need to run the intent links with the other correct way, so any modify to this PR is good, anyone who have full experience to fix this issue can modify this PR.

GertVen:
It would be real nice if we can merge this into the next release of the InAppBrowser. I've had problems with an Ionic app on Androud 11 where I could not login using Auth0, I kept on getting the ERR_UNKNOWN_URL_SCHEMA error, and this PR was the only one that managed to solve the issue.

Tested on an Android 11 emulator

mosabab:
> It would be real nice if we can merge this into the next release of the InAppBrowser. I've had problems with an Ionic app on Android 11 where I could not login using Auth0, I kept on getting the ERR_UNKNOWN_URL_SCHEMA error, and this PR was the only one that managed to solve the issue.
> 
> Tested on an Android 11 emulator

Yes this PR can solve many issues regarding the error : ERR_UNKNOWN_URL_SCHEMA,
All developers are welcome to improve this PR.

ikleiman:
@mosabab I'm facing the same issue, while this PR solves the problem, do you know why does it haven't been merged into the main project after 2 years?

rasmusmilz:
Can anyone help answering why this pr was closed?
I'm in dire need of this, and considering forking to complete in this my code.

PR #927, opened 2022-01-08
feat(ios): explicitly import headers

alekop:
dependencies, instead of relying on PCH files. This is important in Swift projects, where you cannot use prefix headers.

### Platforms affected

iOS

### Motivation and Context
Cannot add plugin to Swift-based project with embedded WKWebView, because it relies on prefix header, which is not supported in Swift projects.

### Description
Explicitly import needed files, so we don't need to rely on PCH files.

### Testing
Built successfully in ObjC and Swift based projects.

### Checklist

### DIFF
--- a/src/ios/CDVInAppBrowserNavigationController.h
+++ b/src/ios/CDVInAppBrowserNavigationController.h
@@ -17,6 +17,7 @@
  under the License.
  */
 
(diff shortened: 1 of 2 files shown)


erisu (approved):
Seems OK to me.

PR #966, opened 2022-07-19
(android) New 'enablethirdpartycookies' option and code clean-ups

fquirin:
### Platforms affected

Android.

### Motivation and Context

- Third party cookies should be configurable and not just simply on by default
- I've done some additional code improvements to better handle resources
- Fixed a type and some potentially confusing, redundant code

### Description

- Third party cookies are often used for tracking and thus should at least be configurable.
- The code for loading drawables kept repeating so I decided to introduce a function to handle it and making it easier to adapt in case of changes. In addition I thought it was useful to handle all required resources as static variables at the top.
- 'hardwarebackbutton' was misspelled, so I fixed that.
- 'showWebPage' redefined some defaults that were already defined at the top which could be very confusing if the default has to be changed in the future.
- Minor code clean-ups for better readability.

### Testing

- Built the app with the new version
- Checked if browser loads as expected
- Changed third party cookie support with new option
- Testes results with `https://www.whatismybrowser.com/detect/are-third-party-cookies-enabled`
- All working as expected :-)

### Checklist


fquirin:
Sorry, I accidentally pushed unwanted changes to the same branch 🤦‍♂️ , will resubmit the PR

PR #967, opened 2022-07-19
(android) New 'enablethirdpartycookies' option and code clean-ups

fquirin:
### Platforms affected

Android.

### Motivation and Context

- Third party cookies should be configurable and not just simply on by default
- I've done some additional code improvements to better handle resources
- Fixed a type and some potentially confusing, redundant code

### Description

- Third party cookies are often used for tracking and thus should at least be configurable.
- The code for loading drawables kept repeating so I decided to introduce a function to handle it and making it easier to adapt in case of changes. In addition I thought it was useful to handle all required resources as static variables at the top.
- 'hardwarebackbutton' was misspelled, so I fixed that.
- 'showWebPage' redefined some defaults that were already defined at the top which could be very confusing if the default has to be changed in the future.
- Minor code clean-ups for better readability.

Note: I noticed there is a lot more potential for optimizing the code. Android Studio itself is already pointing out several simplifications. If there is enough interest I'd consider creating another PR :-).

### Testing

- Built the app with the new version
- Checked if browser loads as expected
- Changed third party cookie support with new option
- Testes results with `https://www.whatismybrowser.com/detect/are-third-party-cookies-enabled`
- All working as expected :-)

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -124,9 +124,9 @@ public class InAppBrowser extends CordovaPlugin {
         CLOSE_BUTTON_CAPTION, TOOLBAR_COLOR, NAVIGATION_COLOR, CLOSE_BUTTON_COLOR, FOOTER_COLOR);
 
     // Resources
-	private static final String TOOLBAR_CLOSE_BUTTON = "ic_action_remove";
-	private static final String TOOLBAR_BACK_BUTTON = "ic_action_previous_item";
-	private static final String TOOLBAR_FORWARD_BUTTON = "ic_action_next_item";
+    private static final String TOOLBAR_CLOSE_BUTTON = "ic_action_remove";
+    private static final String TOOLBAR_BACK_BUTTON = "ic_action_previous_item";
+    private static final String TOOLBAR_FORWARD_BUTTON = "ic_action_next_item";
 
     private InAppBrowserDialog dialog;
     private WebView inAppWebView;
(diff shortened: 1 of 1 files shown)


src/android/InAppBrowser.java
```diff
@@ -799,7 +813,7 @@ public void run() {
 
                 // Toolbar layout
                 RelativeLayout toolbar = new RelativeLayout(cordova.getActivity());
-                //Please, no more black!
+                //"Please, no more black!" <- who said this?!? Black is awesome, especially on OLED! ^^
```
GitToTheHub, line 816: Hi you can remove the complete comment

src/android/InAppBrowser.java
```diff
@@ -626,12 +643,6 @@ private InAppBrowser getInAppBrowser() {
      * @param features jsonObject
      */
     public String showWebPage(final String url, HashMap<String, String> features) {
-        // Determine if we should hide the location bar.
-        showLocationBar = true;
-        showZoomControls = true;
-        openWindowHidden = false;
-        mediaPlaybackRequiresUserGesture = false;
-
```
GitToTheHub, line 634: I'm not sure if `showWebPage` could be called multiple times and the variables are set to the defaults back here. Are you sure, this is not happening here and the defaults can be removed?

GitToTheHub (changes requested):
Some minor changes to do

GitToTheHub:
You would have to add the option `enablethirdpartycookies` to the readme of the Android options.

Issue #706, opened 2020-05-25
Unable to access camera for record video online from inappbrowser


SameerCHS:
Camera is not able to use the camera for recording the video. For example https://webcamtests.com/ even these websites also not able to access the video recording.
I am using ionic 4
camera permission to application is granted(Not sure how to use for in app browser)

Please help me to access the camera

jcesarmobile:
please, when reporting issues provide all the requested data from the bug template, otherwise we can't help you

SameerCHS:
Hi @jcesarmobile, I have updated the reported data in the bug template, Please take a look and revert me ASAP

trackmyapp:
I think I have the same issue.
I added the "NSCameraUsageDescription" and "NSPhotoLibraryUsageDescription" in the config.xml file. So the user gets asked for permission when trying to take a photo or video from an open website in the inappbrowser. If he gives the permission everything works fine, but if he denies the permission in the prompt or in the settings the camera is still being opened, only showing the camera buttons like cancel and a black screen.

If someone knows how to catch that error that would be great! Then we could just show the user a message and switch to the app settings.

In Xcode following error message is shown:
`[Back Camera][com.apple.avfoundation.avcapturedevice.built-in_video:0]>) failed: (Error Domain=AVFoundationErrorDomain Code=-11852 "Cannot use Back Camera" UserInfo={NSLocalizedFailureReason=This app is not authorized to use Back Camera., AVErrorDeviceKey=<AVCaptureFigVideoDevice: 0x1016c77e0 [Back Camera][com.apple.avfoundation.avcapturedevice.built-in_video:0]>, NSLocalizedDescription=Cannot use Back Camera})`

prakash-partho:
I'm also facing same issue as I'm using video conference meeting through Inappbrowser which is not allowing camer and microphone, Is there any solution?

breautek:
I'm reaching here, but even though you may not use them in your main webview/app, your app still needs the permissions. You can try adding the `cordova-plugin-camera` plugin for camera and `cordova-plugin-media-capture` for microphone permissions.

yaminikonduru:
> I'm reaching here, but even though you may not use them in your main webview/app, your app still needs the permissions. You can try adding the `cordova-plugin-camera` plugin for camera and `cordova-plugin-media-capture` for microphone permissions.
> 
[...]

PR #968, opened 2022-07-25
fix(GH-706): (android) Allow permissions requests

joanbat:
### Platforms affected
Android

### Motivation and Context
Android in-app-browser does not allow in-browser camera/webcamera usage (permission denied)
With this change, camera usage will be allowed as long as the app has permissions.
Issue GH-706: https://github.com/apache/cordova-plugin-inappbrowser/issues/706

### Description
Allows permissions requests from ChromeWebView

### Testing
IAB would show permissions denied error in website https://webcam-test.com/
After the changes it successfuly works (requires app's camera permission to be enabled)

### Checklist

### DIFF
--- a/src/android/InAppChromeClient.java
+++ b/src/android/InAppChromeClient.java
@@ -34,6 +34,7 @@ import android.webkit.WebStorage;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import android.webkit.GeolocationPermissions.Callback;
+import android.webkit.PermissionRequest;
 
 public class InAppChromeClient extends WebChromeClient {
 
@@ -45,6 +46,13 @@ public class InAppChromeClient extends WebChromeClient {
         super();
         this.webView = webView;
     }
+    
+    public void onPermissionRequest(final PermissionRequest request) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            request.grant(request.getResources());
+        }
+    }
+
     /**
      * Handle database quota exceeded notification.
      *


asfackahamed (commented):
Your solution is working fine for microphone access 👍

rushabhmshah99:
Spent numerous hours trying to find a solution for this. I can confirm the above change makes it work! Would vote to get it merged asap

rushabhmshah99 (approved):
Spent numerous hours trying to find a solution for this. I can confirm the above change makes it work! Would vote to get it merged asap

ittlsa:
plase confirm this fix.

joanbat:
Yes, the permissions must be declared in AndroidManifest.xml, additionally, permissions must be granted as with this change there is no OS permission popup if permissions have not been allowed.

breautek:
> additionally, permissions must be granted as with this change there is no OS permission popup if permissions have not been allowed.

Ok, so to be clear we have the following use cases:

`Declared In Manifest` means the `<uses-permission>` entry for the permission in question was present in AndroidManifest.
`Permission Granted Previously` means that the permission was granted by some means **outside** of the WebView. Either via another action that triggered the OS permission prompt, or the user used the OS App Permission settings to enable permission manually.
`Permission OK'ed in IAB` means that the permission was granted and the in-app-browser webview was able to utilise the feature.

## Case 1

|Declared In Manifest|Permission Granted Previously|Permission OK'ed in IAB|
|---|---|---|
|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|

Because the `AndroidManifest` contains the permission declaration, and the user has somehow previously granted permission (which has yet to be revoked), the IAB was able to receive the permission and utilise the related features.

## Case 2

|Declared In Manifest|Permission Granted Previously|Permission OK'ed in IAB|
|---|---|---|
|:x:|:x:|:x:|

Because the `AndroidManifest` does not contain the permission declaration, the user would have never been able to have the permission granted previously, and naturally neither would the IAB browser.

## Case 3

|Declared In Manifest|Permission Granted Previously|Permission OK'ed in IAB|
|---|---|---|
|:heavy_check_mark:|:x:|:x:|

The `AndroidManifest` does contain the permission declaration, but the user has yet to trigger a OS permission prompt somehow and does not have the permission currently granted. IAB is unable to trigger a permission prompt, so it will also fail to receive the grant.

Do these case results correctly represent the behaviour as the PR stands now?

joanbat:
Exactly.

breautek:
Awseome.. ok

Personally I feel like this doesn't go far enough to allow permission requests, but I understand it is still an improvement over what the plugin is doing now. Since this can be further improved later, I've written a snippet that I think we could probably include at the end of the `README` file that provides some notes regarding permissions.

```
### Permissions

Some webview features may require additional permissions. The in-app browser does not have the ability to grant additional permissions. However, on Android if the permissions have already been granted, the related featureset will become available. On iOS, permission-locked features may still be unavailable regardless of permission grants.
```

Also, I'm not certain if the iOS claim is accurate. I know this PR doesn't touch iOS, but we should double check iOS behaviour if possible. Unfortunately I'm not in the position to check this myself as I do not have access to a Mac. If we are unable to check the behaviour, then it might be better to scratch the `On iOS, permission-locked features may still be unavailable regardless of permission grants.` line.

Outside of that, I think the PR is good but I'll request further input from a couple of other members who have worked more closely with the IAB than I have.

breautek (approved):

joanbat:
> Also, I'm not certain if the iOS claim is accurate. I know this PR doesn't touch iOS, but we should double check iOS behaviour if possible. Unfortunately I'm not in the position to check this myself as I do not have access to a Mac.

I only tested video, but during my tests I could use webcam and mic access on devices with iOS version 14.3 or higher - That's then WebRTC was introduced in WKWebKit, which is what IAB on iOS uses, as far as I am aware.

With the issue being reported prior to iOS 14.3 release (dec 14th 2020), and that I was able to use permission sensitive features from iOS's IAB during my tests, I would highly suspect it was entirely due to not supporting WebRTC on iOS, rather than having permissions passthrough issues.

breautek:
> > Also, I'm not certain if the iOS claim is accurate. I know this PR doesn't touch iOS, but we should double check iOS behaviour if possible. Unfortunately I'm not in the position to check this myself as I do not have access to a Mac.
> 
> I only tested video, but during my tests I could use webcam and mic access on devices with iOS version 14.3 or higher - That's then WebRTC was introduced in WKWebKit, which is what IAB on iOS uses, as far as I am aware.
> 
> With the issue being reported prior to iOS 14.3 release (dec 14th 2020), and that I was able to use permission sensitive features from iOS's IAB during my tests, I would highly suspect it was entirely due to not supporting WebRTC on iOS, rather than having permissions passthrough issues.

Ok, so we can probably scratch the iOS claim in my snippet.

```
### Permissions

Some webview features may require additional permissions. The in-app browser does not have the ability to grant additional permissions. However, on Android if the permissions have already been granted, the related features will become available.
```

Parag-Tharani:
**@NiklasMerz** could you please approve above PR

kulkarni005:
Please Merge this to master

Meng991003:
I have tried with the above solution, adding the permission in AndroidManifest.xml and granted the access. But I still not be able to open the camera when using IAB. The website is using the input tag <input type="file" id="myFile" name="filename" accept="image/*> to get the input.

StefaniToto:
The fix is working. Thank you

jhlee8804:
Not working on camera.

NiklasMerz:
@breautek If you think this fine feel free to merge. I could create a new release if we think it's a good time to do that.

alexandrampolta:
Solution worked for me , for camera access 
Step 1 : 
cordova plugin add cordova-plugin-android-permissions

With in your index.js

 var permissions = cordova.plugins.permissions;
permissions.requestPermission(permissions.CAMERA, success, error);
function error() {
console.warn('Camera permission is not turned on');
}
 function success( status ) {
  console.log(status);
 }
 
 
 
 Step 2 : 
 cordova plugin remove cordova-plugin-inappbrowser
 
 
 step 3 :
 cordova plugin add https://github.com/apache/cordova-plugin-inappbrowser.git (it's same library but looks like it's updated)
 
 step 4: 
cordova build android --packageType=apk
Build your app normally

vasanthnataraj:
@alexandrampolta camera access works but microphone access is still not working in  inappbrowser.

PR #982, opened 2022-10-12
fix(android): Closes #290: InAppBrowser will not destroy WebView after being…

### LINKED ISSUE #290: InAppBrowser will not close when invoking ref.close()

dtarnawsky:
… closed (by invoking ref.close()) causing memory leaks

### Platforms affected
Android

### Motivation and Context
Fixes [open issue 290](https://github.com/apache/cordova-plugin-inappbrowser/issues/290)

### Description
When a web view is closed its resources are deallocated (Android).

### Testing
Tested by ensuring memory was deallocated and closed web view no longer displayed when inspecting with chrome inspector.

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -536,6 +536,14 @@ public class InAppBrowser extends CordovaPlugin {
                             dialog.dismiss();
                             dialog = null;
                         }
+
+                        if (url.equals(new String("about:blank"))) {
+                            inAppWebView.onPause();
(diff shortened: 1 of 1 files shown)


ath0mas:
Duplicate of #882

PR #983, opened 2022-10-31
Catch/Log regex pattern == null error

bennord:
- log error for null regex pattern
- version bump 5.0.5-dev


bennord:
sorry, wrong branch.

PR #995, opened 2023-03-21
Set webView.inspectable to true on iOS >= 16.4

BBosman:
### Platforms affected

iOS

### Motivation and Context

With the introduction of iOS 16.4 WKWebView instances will no longer be inspectable by default.

Details: [Enabling the Inspection of Web Content in Apps](https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/)

### Description

This PR explicitly sets the inspectability to `YES` again for backwards compatibility, but offers an override option.

Preferably this new setting would be set to the default `NO` or wrapped in an `#ifdef DEBUG`, but that would be a breaking change and needs to wait till a (next) major release.

### Testing

Tested the code in one of our own apps.

### Checklist

### DIFF
--- a/src/ios/CDVWKInAppBrowser.m
+++ b/src/ios/CDVWKInAppBrowser.m
@@ -778,6 +778,14 @@ BOOL isExiting = FALSE;
     [self.view addSubview:self.webView];
     [self.view sendSubviewToBack:self.webView];
     
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
+    // Set "inspectable" to true for backwards compatibility with iOS <= 16.3, but allow an override.
+    // Preferably this would be set to the default "NO" or wrapped in an "#ifdef DEBUG",
+    // but that would be a breaking change and needs to wait till a (next) major release.
+    if (@available(iOS 16.4, *)) {
+        self.webView.inspectable = [settings cordovaBoolSettingForKey:@"InspectableWebview" defaultValue:YES];;
+    }
+#endif
     
     self.webView.navigationDelegate = self;
     self.webView.UIDelegate = self.webViewUIDelegate;


peterknezek:
Hi,
the necessary change for inspectable `WKWebView` was also made in the [apache/cordova-ios](https://github.com/apache/cordova-ios/blob/6e71f5470ed7280c2f8ba4348b0188563ca3c6a4/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m#L224C1-L238C1) but in a little different way:

```swift
#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
    // With the introduction of iOS 16.4 the webview is no longer inspectable by default.
    // We'll honor that change for release builds, but will still allow inspection on debug builds by default.
    // We also introduce an override option, so consumers can influence this decision in their own build.
    if (@available(iOS 16.4, *)) {
#ifdef DEBUG
        BOOL allowWebviewInspectionDefault = YES;
#else
        BOOL allowWebviewInspectionDefault = NO;
#endif
        wkWebView.inspectable = [settings cordovaBoolSettingForKey:@"InspectableWebview" defaultValue:allowWebviewInspectionDefault];
    }
#endif
```

I suggest updating the PR code to match the one used in `apache/cordova-ios`. This will not only improve code consistency within the Cordova ecosystem but also determinate the default value for web on whether it is a debug build or not. In the PR value was always set to YES, which allowed inspection for all cases.

breautek:
> I suggest updating the PR code to match the one used in apache/cordova-ios. This will not only improve code consistency within the Cordova ecosystem but also determinate the default value for web on whether it is a debug build or not. In the PR value was always set to YES, which allowed inspection for all cases.

I agree with this. The rationale behind the way cordova-ios does it is to maintain older behaviour on older iOS devices, which xcode effectively hard-coded the web inspector as enabled for debug builds, and disabled for release builds and I think this PR should follow suite.

src/ios/CDVWKInAppBrowser.m
```diff
@@ -778,6 +778,14 @@ - (void)createViews
     [self.view addSubview:self.webView];
     [self.view sendSubviewToBack:self.webView];
     
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
+    // Set "inspectable" to true for backwards compatibility with iOS <= 16.3, but allow an override.
+    // Preferably this would be set to the default "NO" or wrapped in an "#ifdef DEBUG",
+    // but that would be a breaking change and needs to wait till a (next) major release.
+    if (@available(iOS 16.4, *)) {
+        self.webView.inspectable = [settings cordovaBoolSettingForKey:@"InspectableWebview" defaultValue:YES];;
```
HazzMan2409, line 786: Double semi-colon on here cause compilation failure.

HazzMan2409 (commented):
Comment inline.

HazzMan2409:
Sorry, I pushed a comment and then tried to delete it (I’m not used to GitHub…). Anyway, I tried to apply this PR but noticed that it caused my cordova build to fail. I thought it was the double semi-colon in the change that did it but it also looks like the settings object is not working correctly either.

Issue #1005, opened 2023-06-05
iOS 16.4 (and up) webview can not debug in Safari web inspector


martijnvegmond-cw:
# Bug Report
iOS 16.4 (and up) webview can not debug in Safari web inspector

## Problem
No Web inspection possible for the InAppBrowser window, due to changes in iOS 16.4

### What is expected to happen?
Open a InAppBrowser window and inspects it's content to fix, for example, design issues.

### What does actually happen?
InAppBrowser window of our app is not in the list of webpages in Safari: Develop > *Device* > *App*

## Information
[Enabling the Inspection of Web Content in Apps](https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/)

### Environment, Platform, Device

iOS 16.4 and higher
iPhone and iPad

### Version information
Cordova plugin in app browser v5.0.0
Capacitor v4.8.0

## Checklist
- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

peterknezek:
Hi, I can confirm that in iOS 16.4 and later versions, it prevents InAppBrowser from being inspected in Safari's web inspector by default.

There is an open PR (https://github.com/apache/cordova-plugin-inappbrowser/pull/995) addressing the problem. As a workaround until the library is fixed, you can add Swift code to `src/ios/CDVWKInAppBrowser.m` to enable web inspection.

dylanfpaul:
@jcesarmobile , so how do I actually do a "debug" build to make the application available to the web inspector? I tried running a "debug" cordova build, and then putting the resulting build onto testflight with XCode. When I download the build from TestFlight on my device, connect my device to my computer via a cable, it is still unavailable in the web inspector.

I ran a "debug" cordova build by running `cordova build ios ---debug`

Still can't get it to show up as inspectable in Safari on my computer

breautek:
Assuming you're running cordova-ios 6.3 or later with inappbrowser v6 or later, you need to set the `InspectableWebview` preference.

The preference is documented at https://cordova.apache.org/docs/en/12.x/config_ref/index.html

> Default: depends on Debug/Release build
On iOS 16.4 or later, enables or disables the webview inspector. Defaults to true on debug builds and false on release builds. This preference is ignored on iOS 16.3 and earlier; the inspector is always enabled on debug builds and always disabled on release builds.

[...]

PR #1015, opened 2023-08-24
feat(ios): Make WebView inspectable

jcesarmobile:
use the same code cordova-ios uses for making the WKWebView inspectable, making it inspectable by default in debug builds and respecting the new preference

closes https://github.com/apache/cordova-plugin-inappbrowser/pull/995
closes https://github.com/apache/cordova-plugin-inappbrowser/issues/1005


erisu (approved):
LGTM

PR #1016, opened 2023-08-25
chore: use https urls in tests and README

jcesarmobile:
new cordova apps don't allow http connections by default, so the examples will not work, also it's recommended to use https, so changed the README examples and tests to use https urls.


erisu (approved):
LGTM

PR #1017, opened 2023-08-25
chore: Update SUPPORT_QUESTION.md template

jcesarmobile:
Add GitHub Discussions link
Update slack invite link as current one is down
Change Ionic Slack with Ionic Discord
Remove Phonegap forums since they no longer exist


erisu (approved):
LGTM

Issue #1022, opened 2023-09-20
Android FR: Enable zoom without displaying on-screen zoom controls


reznej:
# Feature Request
Inappbrowser allows zoom functionality (pinch-to-zoom and via on-screen buttons) when the InAppBrowser instance is initialized with the option `zoom` set to `yes`. When using pinch-to-zoom gesture, displaying additional on-screen zoom buttons is not necessary. Therefore, it would be great to enable zooming functionality without displaying the buttons.

## Motivation Behind Feature
- There Is currently no option to enable zoom controls via pinch-to-zoom gesture without showing the on-screen zoom controls
- On-screen zoom controls were deprecated in Android API level 26 - [Source](https://developer.android.com/reference/android/webkit/WebSettings#setDisplayZoomControls(boolean))

## Feature Description
- There should be a separation between enabling zoom-controls (pinch-to-zoom) and additionally showing zoom-buttons to manually adjust zooming.
- It would be great to have two options (zoom and zoomcontrols) when initializing the Inappbrowser instance, so it's up to the developer to display zoom buttons or not.

PR #1023, opened 2023-09-20
feat(android): make zoom on-screen controls configurable

reznej:
### Platforms affected
Android

### Motivation and Context
Closes Issue #1022 

### Description
This change adds an settable option `zoomcontrols` in addition to `zoom` at initialisation of the Inappbrowser instance. This enables a separation between zoom-controls (pinch-to-zoom) and additionally showing zoom-buttons to manually adjust zooming.

### Testing
Ran manual tests with different combinations for parameters `zoom` and `zoomcontrols`.

### Checklist


GitToTheHub (changes requested):

reznej:
Thanks for the review. I addressed all comments.

src/android/InAppBrowser.java
```diff
@@ -631,7 +633,8 @@ private InAppBrowser getInAppBrowser() {
     public String showWebPage(final String url, HashMap<String, String> features) {
         // Determine if we should hide the location bar.
         showLocationBar = true;
-        showZoomControls = true;
+        enableZoom = true;
+        showZoomControls = false;
```
GitToTheHub: The default of `showZoomControls` was `true` before. This should be still the default.

README.md
```diff
@@ -119,7 +119,8 @@ instance, or the system browser.
     - __navigationbuttoncolor__: set to a valid hex color string, for example: `#00ff00`, and it will change the color of both navigation buttons from default. Only has effect if user has location set to `yes` and not hidenavigationbuttons set to `yes`.
     - __toolbarcolor__: set to a valid hex color string, for example: `#00ff00`, and it will change the color the toolbar from default. Only has effect if user has location set to `yes`.
     - __lefttoright__: Set to `yes` to swap positions of the navigation buttons and the close button. Specifically, navigation buttons go to the right and close button to the left. Default value is `no`.
-    - __zoom__: set to `yes` to show Android browser's zoom controls, set to `no` to hide them.  Default value is `yes`.
+    - __zoom__: Enables pinch-to-zoom gesture. Set to `no` to disable it. Default value is `yes`.
+    - __zoomcontrols__: set to `yes` to show Android browser's zoom controls, set to `no` to hide them. Default value is `no`.
```
GitToTheHub: `zoomcontrols` should be yes by default, because `zoom` enabled them by default also.

src/android/InAppBrowser.java
```diff
@@ -128,7 +129,8 @@ public class InAppBrowser extends CordovaPlugin {
     private EditText edittext;
     private CallbackContext callbackContext;
     private boolean showLocationBar = true;
-    private boolean showZoomControls = true;
+    private boolean enableZoom = true;
+    private boolean showZoomControls = false;
```
GitToTheHub: `showZoomControls` was `true` before. This should be still the default.

GitToTheHub (changes requested):
The option `showZoomControls` should be true as default. Otherwise this looks good to me.

reznej:
> The option `showZoomControls` should be true as default. Otherwise this looks good to me.

Just for clarification: should both pinch-to-zoom and showZoomControls default to true?

GitToTheHub:
Yes, because this was the default for Android before.

reznej:
Done. Thank you!

src/android/InAppBrowser.java
```diff
@@ -946,7 +953,11 @@ public boolean onShowFileChooser (WebView webView, ValueCallback<Uri[]> filePath
                 WebSettings settings = inAppWebView.getSettings();
                 settings.setJavaScriptEnabled(true);
                 settings.setJavaScriptCanOpenWindowsAutomatically(true);
-                settings.setBuiltInZoomControls(showZoomControls);
+                // Enables built-in zoom mechanisms, which are on-screen zoom controls and pinch-to-zoom.
+                // The on-screen zoom controls have to be enabled/disabled separately after.
+                // The on-screen zoom controls are deprecated since Android API Level 26 (Android 8).
```
GitToTheHub: Please change the comment to:

```java
// setBuiltInZoomControls enables pinch-to-zoom and also the on-screen zoom controls
// The zoom controls have to be disabled separately by setDisplayZoomControls
// This is the recommended way by Google
```

src/android/InAppBrowser.java
```diff
@@ -946,7 +953,11 @@ public boolean onShowFileChooser (WebView webView, ValueCallback<Uri[]> filePath
...
+                settings.setBuiltInZoomControls(enableZoom);
+                settings.setDisplayZoomControls(showZoomControls);
```
GitToTheHub, line 963: Please add a comment here:

```java
// Enable/Disable on-screen zoom controls.
// These are deprecated since Android API Level 26 (Android 8).
// Google recommends to disable them
```

README.md
```diff
@@ -119,7 +119,8 @@ instance, or the system browser.
     - __navigationbuttoncolor__: set to a valid hex color string, for example: `#00ff00`, and it will change the color of both navigation buttons from default. Only has effect if user has location set to `yes` and not hidenavigationbuttons set to `yes`.
     - __toolbarcolor__: set to a valid hex color string, for example: `#00ff00`, and it will change the color the toolbar from default. Only has effect if user has location set to `yes`.
     - __lefttoright__: Set to `yes` to swap positions of the navigation buttons and the close button. Specifically, navigation buttons go to the right and close button to the left. Default value is `no`.
-    - __zoom__: set to `yes` to show Android browser's zoom controls, set to `no` to hide them.  Default value is `yes`.
+    - __zoom__: Enables pinch-to-zoom gesture. Set to `no` to disable it. Default value is `yes`.
+    - __zoomcontrols__: set to `yes` to show Android browser's zoom controls, set to `no` to hide them. Default value is `yes`.
```
GitToTheHub: Please change the documentation of `zoomcontrols`:

```md
- __zoomcontrols__: set to `yes` to show Android browser's zoom controls, set to `no` to hide them. Default value is `yes`. The zoom controls are deprecated since Android API Level 26 (Android 8). Google recommends to disable them.
```

GitToTheHub (changes requested):
Hi, thanks for the changes. The code comment for `settings.setBuiltInZoomControls` should be changed, because I didn't like the wording I choosed before after re-thinking it. `settings.setDisplayZoomControls` should get its own comment. And `zoomcontrols` should get the note, that they are deprecated since Android 8 and Google recommends to disable them. I added everything in the review.

src/android/InAppBrowser.java
GitToTheHub: I mean add the comment to `settings.setDisplayZoomControls(showZoomControls);`. I selected the wrong range.

GitToTheHub:
Thanks I will let now some checks run. Some are always failing, so don't wonder about them.

GitToTheHub:
@dpogue @breautek @erisu The zoom controls can now be configured separately by this PR and Google recommends to disable them, since they are deprecated since Android 8. The default for this plugin was before, that they are enabled. Should this PR disable them by default? Would this be a breaking change, or this still ok for a minor update? The deprecation and recommended setting is documented on [WebSettings.setDisplayZoomControls](https://developer.android.com/reference/android/webkit/WebSettings#setDisplayZoomControls(boolean)). I think it's the right way to disable them by default since Google recommends it.

dpogue:
That would probably be considered a breaking change, especially since it's user visible

GitToTheHub:
Ok than it's a breaking change. Do you would also consider disabling the controls by default since Google recommends it to do so?

dpogue:
That seems reasonable to do in the next major

GitToTheHub:
There were already a lot of changes on this plugin, but would you consider to first make a minor, before make a major?

dpogue:
I don't usually pay close attention to what's happening on the plugins, so I'm not sure the scope of the fixes and stuff; but as long as there are no breaking changes, then it would be fine to do a minor, and then include the zoom change in a major after that.

The question of whether it's worth doing a minor and then a major vs just doing a major is largely a tradeoff of how badly the fixes are needed for the current version and how much effort will be involved in doing multiple releases.

GitToTheHub:
Thanks for your thoughts. I think it's ok to make a minor first and change this in the next major.

GitToTheHub:
@reznej Thanks for the PR

PR #1033, opened 2023-11-04
chore: bump to next major release 6.0.0

NiklasMerz:
### Motivation and Context

Prepare for next release

See list https://lists.apache.org/thread/p2ycdfv4tf0vnt1dbkv1160o5xz5knns

### Checklist


jcesarmobile (changes requested):
In package.json this has to be updated to 7.0.0 or plugin will fail to update
https://github.com/apache/cordova-plugin-inappbrowser/blob/839e2e2f36f838de91d3de69fb08d7192aecd736/package.json#L45

NiklasMerz:
> In package.json this has to be updated to 7.0.0 or plugin will fail to update
> 
> https://github.com/apache/cordova-plugin-inappbrowser/blob/839e2e2f36f838de91d3de69fb08d7192aecd736/package.json#L45

I updated cli & ios to last major release. Is that correct? Android is not needed, yet I guess from the commits in master right?

jcesarmobile:
I don't think we need them to be latest, cordova-ios 6 supports the same iOS versions as cordova-ios 7. And for the CLI I'm not aware of any reason to bump it, maybe because of nodejs support?
A `cordova-android` entry is needed, in geolocation plugin it was set to `>=10.0.0` recently.
https://github.com/apache/cordova-plugin-geolocation/blob/master/package.json#L34-L37

NiklasMerz:
That sounds good to me. I set the same requirements and kept cli from the previous version. Thanks for the help @jcesarmobile. I never really dug into these requirements before.

jcesarmobile (approved):

PR #1067, opened 2024-08-28
fix(iOS): IAB not showing up in apps using UIScenes

dpogue:
### Platforms affected
iOS

### Motivation and Context

When a Cordova app uses the UIScene APIs introduced in iOS 13, the temporary window created for the IAB ViewController is not associated with that scene and does not appear on screen.

### Description

On iOS 13 and newer, check if the CDVViewController's owner window has an associated scene. If yes, use that when constructing the temporary UIWindow for the IAB view.

### Testing

Tested manually using mobilespec in a simulator with cordova-ios @ 7.1.1.
Tested manually using mobilespec in a simulator with cordova-ios @ 8.0.0-dev.

### Checklist


erisu (approved):
LGTM

Test both iOS 18 and iOS 17.x simulators.

PR #1073, opened 2024-09-22
feat(android): add page loading progress bar

### LINKED ISSUE #1076: Add Loading Indicator (Progress Bar) in InAppBrowser for Android
# Feature Request

## Motivation Behind Feature

Currently, there is no visual feedback indicating that content inside the InAppBrowser is loading. Users experience a blank screen until the webpage fully loads, refer this [(https://github.com/apache/cordova-plugin-inappbrowser/issues/1075)].  This could lead to confusion or a perception of slow performance. Adding a loading indicator (such as a spinner) would enhance the user experience.

## Feature Description

Add a ProgressBar or Spinner:

When the InAppBrowser starts loading a webpage, display a ProgressBar (spinner) in the center of the screen.
Once the webpage finishes loading, hide the ProgressBar.
The feature is dependent on the web content load time. If there is an error in loading the page, the progress bar may remain visible, so error handling must be considered.

## Alternatives or Workarounds

Currently, there is no easy way to visually indicate that the InAppBrowser is loading content. A custom implementation of this loader using native code, as suggested above, could solve the issue.

One workaround is to hide the browser and show it in the loadstop event, but due to this user stays on the native page for a long time and considers that no action is happening. This is because for iOS we have a loading indicator/progress bar.
So, we want the same behaviour for Android also.

MaximBelov:
### Platforms affected
Android

### Motivation and Context
There is no visual feedback while a page is loading in the InAppBrowser on Android, so the user just sits in front of an empty view until the page renders and assumes nothing is happening. iOS already shows a loading indicator, this brings Android in line with it.

The same thing was reported from the other side in #1075: opening a link gives a black screen (white on emulators, it follows the theme) for 2-3 seconds before the page appears. The suggested `hidden` + `loadstop` workaround hides the blank frame but leaves the user staring at the previous screen with no feedback instead, and the reporter said they still caught a glimpse of black. This PR doesn't remove that blank frame, it just means there is a progress indicator moving while it happens.

Closes #1076
Refs #1075

### Description
Adds a horizontal `ProgressBar` to the browser layout, between the toolbar and the WebView. It is driven by `onProgressChanged` on the chrome client and is set to `GONE` once loading reaches 100%, so it takes no space while you are actually reading the page.

It can be turned off with a new `hideprogressbar=yes` option, following the same pattern as `hideurlbar` and `hidenavigationbuttons`. Default is `no`, so the bar is shown.

Rebased on current master, so this now sits on top of the `onCreateWindow` changes rather than conflicting with them.

### Testing
Tested by hand on Android against a slow page: the bar fills as the page loads and disappears when it finishes, with no leftover gap in the layout. Checked with `location=yes`, `location=no` and `footer=yes`, and with `hideprogressbar=yes` to confirm nothing is added to the layout in that case.

`npm test` passes, and the plugin compiles against cordova-android 14 / SDK 35.

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -60,6 +60,7 @@ import android.widget.EditText;
 import android.widget.ImageButton;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
+import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 
@@ -916,12 +917,23 @@ public class InAppBrowser extends CordovaPlugin {
                 View footerClose = createCloseButton(7);
                 footer.addView(footerClose);
 
+                // Progress of loading a page
+                ProgressBar progressBarLoadingPage = new ProgressBar(webView.getContext(), null, android.R.attr.progressBarStyleHorizontal);
+                progressBarLoadingPage.setLayoutParams((new LinearLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT)));
+
                 // WebView
                 inAppWebView = new WebView(cordova.getActivity());
+                inAppWebView.setBackgroundColor(Color.WHITE);
+                inAppWebView.addView(progressBarLoadingPage);
                 inAppWebView.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
                 inAppWebView.setId(Integer.valueOf(6));
                 // File Chooser Implemented ChromeClient
                 inAppWebView.setWebChromeClient(new InAppChromeClient(thatWebView) {
+                    public void onProgressChanged(WebView view, int progress)
+                    {
+                        progressBarLoadingPage.setProgress(progress);
+                        progressBarLoadingPage.setVisibility(progress == 100 ? ProgressBar.GONE : ProgressBar.VISIBLE);
+                    }
                     public boolean onShowFileChooser (WebView webView, ValueCallback<Uri[]> filePathCallback, WebChromeClient.FileChooserParams fileChooserParams)
                     {
                         LOG.d(LOG_TAG, "File Chooser 5.0+");


GitToTheHub:
Hi,

since you gave maintainers the possibility to make changes on your PR, I changed this PR to be inline with iOS.

iOS shows a spinner and has the option `hidespinner` to disable this. I added a spinner instead of a progress bar, which can also be disabled by the option `hidespinner`. The spinner looks like this on Android 16:



The background is indeed black here. On Android 10 it is white. So the background view should be set as white by default like iOS.

If no one has any objections, this could be merged.

MaximBelov:
Thanks for taking it forward. I built test apps and measured both implementations instead of arguing from the code. Summary below — no objection to merging your spinner.

**The bar is not worth defending.** Fill measured from pixels, Android 15:

- Slow first byte (the #1075 / #1076 case): the bar is up for the whole 7s stall, but frozen at exactly 10% across 41 consecutive samples. Chromium reports progress 10 while awaiting the document, then jumps to 100.
- Fast document, slow subresources: climbs 10 → 20 → 80% in about 400ms, then disappears while the page keeps loading for another ~12s.

So determinate buys nothing on screen. The one thing I'd still mildly prefer is placement — a strip under the toolbar doesn't cover content the user is reading on later in-page navigations — but that's a preference, not a defect.

**Your hide condition is fine.** I earlier claimed aborted loads leave the spinner running. Measured, that is wrong: attachment downloads, a cancelled SSL handshake and a `tel:` hand-off all hide it within ~1.5s, because WebView still delivers `progress == 100` for the aborted navigation. No extra hide calls needed.

**Separate issue those runs turned up**, not caused by this PR: on a cancelled SSL handshake or an attachment URL, the browser is left on a blank page with the URL still in the toolbar and no error shown — and the attachment is never fetched, since `DownloadListener` only emits the JS `download` event. Happy to file it.

Test apps, pinned to the relevant commits, with screenshots and the servers used:

- https://github.com/MaximBelov/cordova-iab-pr1073-progressbar-demo
- https://github.com/MaximBelov/cordova-iab-pr1167-backgroundcolor-demo

On the background colour you raised: #1167 covers it. Worth knowing the default follows the device theme — in light mode the empty frame is already white, so it only shows black in dark mode.

GitToTheHub:
I created a new PR #1169 which superseds this PR. A progress bar is a more modern solution. Another PR will add a progress bar to iOS also.

PR #1127, opened 2026-02-25
doc(ios): constraint warnings on iOS 26 and using `initWithBarButtonSystemItem`

GitToTheHub:
- Add a note, that using `initWithBarButtonSystemItem` gives constraint warnings on iOS 26, which is due to a iOS bug

### Checklist


dpogue (commented):
My concern with adding comments like this (especially when the code itself is correct) is that 5 years from now there will be confusion when someone tries to refactor the code and doesn't know if there's a bug here or not.

GitToTheHub:
I reworded the note, is this fine, or does it confuse?

PR #1138, opened 2026-04-24
feat(ios): use `UIBarButtonSystemItemClose` since iOS 26

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

- Since iOS 13 it's possible to use `UIBarButtonSystemItemClose` instead of `UIBarButtonSystemItemDone` which correspond perfectly to a close button. Since `UIBarButtonSystemItemClose` looks akward on iOS 18 and older, because an X inside a round button is shown, instead of a text, the close button is only applied since iOS 26.

### Checklist


GitToTheHub:
Is this fine for everyone to merge?

PR #1143, opened 2026-05-05
fix(ios): set callbackId only on target=_blank, update readme

GitToTheHub:
### Platforms affected

- iOS

### Motivation and Context

I want to open a link in the external browser, which is clicked in the in-app browser, by listening to the event `beforeload` and open the link by `cordova.InAppBrowser.open(externalUrl, '_system')`. After doing that, the `beforeload` event is not triggered anymore because the in-app browser reference has changed by `open(externalUrl, '_system')`. Since `_system` is used and the in-app browser just forwards the call, it's is not necessary to change the reference. Also the documentation states that the reference to the `InAppBrowser` window is set, when the target is set to `'_blank'` only.

On Android, non-whitelisted URLs also fall back to the `InAppBrowser` when the target is set to `'_self'`, which is not the case on iOS. This case was updated in the README.

### Testing

Tested successfully in an App with the event `beforeload` and opening a specific link external:

```js
const inAppBrowserRef = cordova.InAppBrowser.open(url, "_blank", "beforeload=yes");

inAppBrowserRef.addEventListener("beforeload", (inAppBrowserEvent, callback) => {
      // Open clicked link external
      if (inAppBrowserEvent.url == "https://myExternalUrl.com") {
        cordova.InAppBrowser.open(inAppBrowserEvent.url, "_system");
      } else {
        // Load URL in in-app browser
        callback(inAppBrowserEvent.url);
      }
});
```

On Android only the first page load in the in-app browser will fire `beforeload`. After this the event will not be fired again, e.g. when a link is clicked, which seems to be a bug on Android and has nothing to do with this PR.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,13 +92,13 @@ instance, or the system browser.
 
     var ref = cordova.InAppBrowser.open(url, target, options);
 
-- __ref__: Reference to the `InAppBrowser` window when the target is set to `'_blank'`. _(InAppBrowser)_
+- __ref__: Reference to the `InAppBrowser` window when the target is set to `'_blank'`. _(InAppBrowser)_ On Android, non-whitelisted URLs also fall back to the `InAppBrowser` when the target is set to `'_self'`.
 
 - __url__: The URL to load _(String)_. Call `encodeURI()` on this if the URL contains Unicode characters.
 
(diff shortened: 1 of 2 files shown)


GitToTheHub:
If no one has any objections I would like to merge it

PR #1144, opened 2026-05-05
fix(ios): don't call `beforeload` on POST and `beforeload=yes`

GitToTheHub:
### Platforms affected

- iOS

### Motivation and Context

- If the option `beforeload=yes` was set and a POST-Request was made, `beforeload` was fired which was wrong, because POST-Request are not supported currently for beforeload
- Improve dectection for `isTopLevelNavigation`: Using `targetFrame` makes top-level detection based on actual frame intent, not URL coincidence, so the interception logic is more correct and robust.
- Generated-By: GPT-5.3-Codex

### Checklist


GitToTheHub:
If no one has any objections I would like to merge it

PR #1146, opened 2026-05-06
fix(android): `beforeload` event fires only once

GitToTheHub:
### Platforms affected

- Android

### Motivation and Context

- When beforeload=yes was set, the beforeload event would only fire on the initial page load, then stop firing for subsequent navigations.
- Route onCreateWindow navigation through the active InAppBrowser client so beforeload and URL overrides are consistently applied
- Ignore transient about:blank bootstrap URLs from new-window flow to avoid breaking the visible page
- Re-arm waitForBeforeload after page finish and load error so interception keeps working after the first allowed navigation
- Generated-By: GPT-5.3-Codex

### Checklist


GitToTheHub:
If no one has any objections I would merge it

PR #1147, opened 2026-05-06
fix(android): change callback context only on non-whitelisted urls or target `_blank`

### LINKED ISSUE #752: Events are not isolated between multiple instances of IAB
# Bug Report

## Problem

If multiple instances of IAB are created, only the events from the last created IAB instances are called back.

Consider, we create one InAppBrowser-

```javascript
const inAppBrowserRef1 = cordova.InAppBrowser.open('http://example.com', '_blank', 'location=yes');

inAppBrowserRef1.addEventListener('loadstart', loadStartCallBack);
inAppBrowserRef1.addEventListener('exit', loadExitCallBack);
inAppBrowserRef1.addEventListener('message', messageCallBack);
```

All the callback works fine. Now, while this IAB is open and we launch another IAB with target as `_system`-

```javascript
const inAppBrowserRef2 = cordova.InAppBrowser.open('http://google.com', '_system');

inAppBrowserRef2.addEventListener('loadstart', loadStartCallBack);
inAppBrowserRef2.addEventListener('exit', loadExitCallBack);
inAppBrowserRef2.addEventListener('message', messageCallBack);
```

After this code is executed, all the event listeners on `inAppBrowserRef1` stop working and instead they start to execute on `inAppBrowserRef2` listeners for the first IAB instance.

Event callbacks on `inAppBrowserRef1` should work irrespective of other IAB i.e. `inAppBrowserRef2`.

## Information

Came to know about this problem while implementing message listeners and I was able to listen the `message` events only once as described here https://gist.github.com/sagrawal31/9004198ae56a58ff235ce0f1dd62362c

### Command or Code

Already shared the code above.

### Environment, Platform, Device, Version information

- Cordova- 9.0.0 (cordova-lib@9.0.1)
- cordova-android- 8.1.0 & 9.0.0
- Android OS- Samsung UI 2.0 (Android 10)
- Plugin version- ^4.0.0
- OS- macOS Catalina
- Tested on a fresh/empty Cordova app
- Android Studio- 4

## Checklist

GitToTheHub:
### Platforms affected

- Android

### Motivation and Context

- Opening a link externally via `_system` in a `beforeload` callback, would break the `beforeload` callback for the remaining InAppBrowser session.
- The callback context was always changed, when `open` was called, no matter what should be loaded. Now it will only be set for non-whitelisted urls and target `_blank`, like the documentation already states.
- It is not applied anymore for:
  - _system
  - _self + whitelisted URL
  - _self + tel:
- Fixes https://github.com/apache/cordova-plugin-inappbrowser/issues/752
- Generated-By: GPT-5.3-Codex

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -169,11 +169,6 @@ public class InAppBrowser extends CordovaPlugin {
                 t = SELF;
             }
             final String target = t;
-            // Keep the existing callback channel when opening _system from inside
-            // an active InAppBrowser instance (e.g. from beforeload handling).
-            if (!SYSTEM.equals(target)) {
-                this.callbackContext = callbackContext;
-            }
             final HashMap<String, String> features = parseFeature(args.optString(2));
 
             LOG.d(LOG_TAG, "target = " + target);
@@ -240,6 +235,7 @@ public class InAppBrowser extends CordovaPlugin {
                         // load in InAppBrowser
                         else {
                             LOG.d(LOG_TAG, "loading in InAppBrowser");
+                            InAppBrowser.this.callbackContext = callbackContext;
                             result = showWebPage(url, features);
                         }
                     }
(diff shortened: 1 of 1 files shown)


GitToTheHub:
If no one has any objections, I would merge it

PR #1148, opened 2026-05-06
fix(android): place footer below WebView instead of overlaying content

### LINKED ISSUE #294: Android - Footer Overlaps WebView
On Android, when setting `footer` to `yes`, the footer bar overlaps the WebView content, causing it to cover up the bottom slice of the WebView.

I was able to fix the issue by forking the library and using `setMargins` on the WebView's `LinearLayout.LayoutParams` equal to the height of the footer, but I only did that because I could not get any other way to work so I am not sure if this is the best way to fix the problem.

My fix is below.

FROM:

    inAppWebView = new WebView(cordova.getActivity());
    inAppWebView.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));

TO:

    int footerSize = this.dpToPixels(44);

    inAppWebView = new WebView(cordova.getActivity());

    LinearLayout.LayoutParams webViewLayoutParams = new LinearLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT);

    if (showFooter) {
        webViewLayoutParams.setMargins(0, 0, 0, footerSize); // Adding margin the same size as the footer
    }

    inAppWebView.setLayoutParams(webViewLayoutParams);

GitToTheHub:
### Platforms affected

- Android

### Motivation and Context

- On Android, enabling footer=yes adds the footer as a child of the WebView container, so it is rendered on top of the page content. This causes the bottom of the page to be obscured unless the loaded site adds manual bottom padding.
- This change moves the footer into the main vertical layout flow below the WebView and gives the WebView container a weighted height. That way, the footer keeps its fixed height while the WebView uses the remaining space, and page content is no longer covered by the native footer.
- Set the WebView container to LinearLayout.LayoutParams(MATCH_PARENT, 0, 1.0f).
- Stop adding the footer as an overlay inside webViewLayout.
- Add the footer directly to the main vertical layout with its own fixed toolbar-height layout params.
- Fixes: https://github.com/apache/cordova-plugin-inappbrowser/issues/294
- Generated-By: GPT-5.4

### Checklist

### DIFF
--- a/src/android/InAppBrowser.java
+++ b/src/android/InAppBrowser.java
@@ -1047,12 +1047,14 @@ public class InAppBrowser extends CordovaPlugin {
 
                 // Add our webview to our main view/layout
                 RelativeLayout webViewLayout = new RelativeLayout(cordova.getActivity());
+                webViewLayout.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.MATCH_PARENT, 0, 1.0f));
                 webViewLayout.addView(inAppWebView);
                 main.addView(webViewLayout);
 
                 // Don't add the footer unless it's been enabled
                 if (showFooter) {
-                    webViewLayout.addView(footer);
+                    footer.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.MATCH_PARENT, this.dpToPixels(TOOLBAR_HEIGHT)));
+                    main.addView(footer);
                 }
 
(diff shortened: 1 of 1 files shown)


GitToTheHub:
If no one has objections, I would merge this

PR #1154, opened 2026-05-20
chore!: raise minimum Cordova dependencies to the penultimate version, bump plugin major version

GitToTheHub:
### Platforms affected

- Android
- iOS

### Motivation and Context

- It's hard to test if old cordova dependencies would still work with the plugin. For e.g. if cordova-android 10 is supported by a plugin, changes would have to test against cordova-android 10, 11, etc.
- To make plugins easier to maintain it's better to keep the requirements not to wide. Also old code could be removed.
- Bump plugin version from 6.1.0-dev to 7.0.0-dev
- Changed versions: -- cordova-android from 10.0.0 to 14.0.0 -- cordova-ios from 6.2.0 to 7.0.0 -- cordova from 9.0.0 to 12.0.0

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cordova-plugin-inappbrowser",
-  "version": "6.1.0-dev",
+  "version": "7.0.0-dev",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


GitToTheHub:
The community decided against raising the minimum Cordova dependencies to the penultimate version to simplify development. The dependencies should only be raised if the code needs this to do.

Issue #1075, opened 2024-10-16
Black Screen gets pop up for 2-3 seconds when we open any link using in-app-browser for Android. 


Akash-Tapadia-DFT:
# Bug Report

## Problem
We are getting a black screen for around 2 or 3 seconds and then the page starts loading in Android. We are able to reproduce for all Android versions, physical as well as emulators.

Note : For emulators you may get a white screen instead of black.

### What is expected to happen?
Black/White screen should not appear. Flow should be similar to ios.

### What does actually happen?
We get a black/white screen for around 2-3 seconds for Android.

### Command or Code

`inappbrowser.create("<any_url>", '_blank', {})`
e.g url = https://github.com

### Environment, Platform, Device

Platform  : Android

Device : Reproducible on all Android devices

### Version information

Environment Details
ionic        :  6.20.1
cordova   :  10.0.0
node        :  16.14.2
inappbrowser plugin version : 6.0.0

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

breautek:
This sounds more like a slow network/device issue rather than a bug.

The black (or white screen, depending on the configured theme) is generally the about:blank page which is the default page for when there is no page loaded in the webview.

Using `cordova.InAppBrowser.open('https://github.com', '_blank', 'location=yes');` (`inappbrowser.create` is an API that doesn't exist at the cordova plugin level), I see the black screen for approximately 1 second when running on a simulator.

If you want to hide this loading, you can utilise the `hidden` flag so that the IAB isn't immediately visible to the user (but it will still load the page). Then you can utilise the `loadstop` / `loaderror` [listeners](https://github.com/apache/cordova-plugin-inappbrowser?tab=readme-ov-file#inappbrowseraddeventlistener) to show the IAB or display some kind of error feedback.

Akash-Tapadia-DFT:
Thank you for your response @breautek 

I have tried using the flag 'hidden'
And in the loadstop event I am calling the show method for the inappbrowser created before.
Even with this I am getting a glimpse of black screen.

PR #1167, opened 2026-08-12
feat(android): add backgroundcolor option

### LINKED ISSUE #501: Is it possible to set background color of web view?
# Feature Request

## Motivation Behind Feature
When use in-app-browser to view am image, the background color will show in white.

## Feature Description
set background color of web view.

MaximBelov:
### Platforms affected
Android

### Motivation and Context
There is no way to set the browser background on Android.

It shows up most when opening the browser: you get the device theme background for a second or two until the first page paints. Black on devices, white on emulators. That is what #1075 was about. The `hidden` + `loadstop` workaround from that thread hides the browser rather than fixing the colour, and the reporter said they still saw the black.

Same missing option in #501, open since 2019, for pages that don't paint their own background such as a plain image.

iOS sets `webView.backgroundColor` to white and `opaque = YES` in `CDVWKInAppBrowser.m`, so the two platforms don't match today.

Closes #501
Refs #1075

### Description
New `backgroundcolor` option, same hex format as `toolbarcolor` and `footercolor`.

It is set on the root `LinearLayout` and on the WebView. The root layout is the one that removes the flash: the WebView draws nothing until it has painted its first frame, so until then you are seeing the window background through an unpainted layout. Setting it on the WebView alone did nothing when I tried that first.

`BACKGROUND_COLOR` also needs to be in `customizableOptions`, otherwise `parseFeature` rewrites the value to `yes` and `Color.parseColor` throws.

Nothing changes if you don't pass the option. I didn't default it to white to match iOS, since that changes how every existing app looks, but say the word if you would rather have it that way round.

No overlap with #1073, either can go in first.

### Testing
Builds against cordova-android 14 / SDK 35, `npm test` clean.

Draft because I haven't put it on a device yet. Needs `backgroundcolor=#ffffff` on a dark themed device to confirm the flash is gone, plus one run without the option to confirm nothing moved.

### Test app

A standalone Cordova app that builds this change and reproduces it on a device, with screenshots:

https://github.com/MaximBelov/cordova-iab-pr1167-backgroundcolor-demo

Worth knowing before reviewing: the default empty frame follows the device theme. In light mode it is already white, so the option reads as a no-op there — the captures with and without it come out byte-identical. The black flash this removes only appears in dark mode.

### Checklist


GitToTheHub:
Hi,

you wrote:

> iOS sets webView.backgroundColor to white and opaque = YES in CDVWKInAppBrowser.m, so the two platforms don't match today.

This should be done also for Android by default. Than there is no need for setting a custom background color. If iOS and Android should support dark mode. This would have to be handled extra.

GitToTheHub:
I tested the in-app browser loading process on an Android 16 emulator and the background was black by default though the emulator was in light mode. On an Android 10 emulator it was white by default. To be inline with iOS it should be white by default. I think an option for setting the background color is unnecessary.

MaximBelov:
Test app so this can be checked on a device, pinned to this branch with screenshots:
https://github.com/MaximBelov/cordova-iab-pr1167-backgroundcolor-demo

Interesting — I get a different default from you on the same version. Android 16, Pixel 9a emulator in light mode: the empty frame is white, and captures with and without `backgroundcolor=#FFFFFF` come out byte-identical. It is black only when the emulator is in dark mode. You saw black in light mode on 16, and white on 10.

That divergence looks like the argument for the option rather than against it. The default depends on Android version, device and the host app's theme, so an app that needs a predictable frame can't rely on it. Hardcoding white would also change behaviour for every existing app, whereas the option changes nothing unless it is passed.

Also fixed a bug this surfaced: the colour used to stick across opens — once set, a later open without the option kept it for the life of the process. `2b897e7d` assigns the field unconditionally. Verified on a device: amber open, then an open with no option measures `rgb(255,255,255)` instead of `rgb(255,204,0)`.

PR #1169, opened 2026-08-13
feat(android): add progress bar

GitToTheHub:
### Platforms affected

Android

### Motivation and Context

- PR #1073 added the same implementation. The PR got a little bit noisy, so this is why this PR was created.
- The progress bar can by hidden by `hideloadingindicator`. The name is hold generic, so it don't presumes a spinner or progress bar.

### Description

- The progress bar is enabled by default but can be hidden by the option `hideloadingindicator`.
- Superseds #1073

Generated-By: GPT-5.6 Terra

### Sample

Android 10

### Checklist

- [ ] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

PR #1170, opened 2026-08-13
feat(ios): switch from spinner to progress bar

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

- PR #1169 implemented a progress bar for Android. Today it's more common to use a progress bar instead of a spinner to indicate loading of web page as it is the default for the Safari app. This aligns it with Android and use the same option `hideloadingindicator` as for Android. The old option `hidespinner` is still supported and will be an alias for `hideloadingindicator`.

### Sample

iOS 26.5 iPhone 17 Pro Simulator

### Checklist


GitToTheHub:
If no one has any objections, I would merge this

